PR #1, opened 2015-11-02
Fixed incorrect downtime severity string comparison

Tom-Byrne:
Incorrect capitalisation of the severity string comparison caused the 'warning' downtime option to always be selected when editing a downtime.


Tom-Byrne:
Messed up this pull request, trying again!

PR #2, opened 2015-11-11
Add_downtime SL6/Chrome fix

Tom-Byrne:
Fix for bug reported here:

https://ggus.eu/?mode=ticket_info&ticket_id=117404

where a user was unable to add a downtime in an older version on chrome, due to the inline JS validation function not being fired and enabling the "add downtime" button.

The inline JS validation function was originally called from the `<form onChange="validate()">`, which apparently is not the right way to do it as <input> changes are not meant to 'bubble' up through the form, and apparently don't in the particular version of chrome the user was using (Chrome Version 27.0.1453.110).

 I've replaced the relevant onChanges with a few jQuery `.change()` events, and a bootstrap event for the datetimepickers.


Tom-Byrne:
At JK's suggestion, I changed the form `onchange` handling on the `edit_downtime.php` page too, as it was handled in a similar fashion, but the bug was not visible due to `validate()` also being called on `onKeyUp` for some reason. I think it's a lot simpler now. I've also added some comments, and fixed a bug I introduced where the datepickers were not calling `validate()`.

PR #3, opened 2015-11-19
Site/endpoint/service property delete checkboxes

Tom-Byrne:
Adding the ability to select and delete multiple properties at the same time


Tom-Byrne:
This is ready for review and testing.

PR #7, opened 2015-12-15
Properties batch upload feature

Tom-Byrne:


Tom-Byrne:
This is ready for testing

PR #17, opened 2016-03-09
Added code to allow paging by default to PI method to GetService

Tom-Byrne:


davidmeredith:
fixed by DM/TB on dev branch

PR #27, opened 2016-06-15
Whitespace - tabs to spaces and trailing spaces

GRyall:
- Replaces leading tabs with 4 spaces each on 8910 lines of 211 .php files
- Replaces leading tabs with ~~4 spaces~~ 2 spaces each on 170 lines of 5 html files
- Removes trailing spaces from 11,912 lines of 92 .php files
- Edit: Removes trailing tabs from 290 lines in 88 files

The commands used are in the commit description for each commit. Appending '?w=1' to the URL of each commit will confirm only white space changes. The unit tests still pass.


jrha:
Unfortunately this isn't quite complete, your first command:
```bash
for file in $(find . -type f -name \*.php | grep -v "^./vendor" | grep -v "^./lib/Doctrine/compiledEntities/"); do
    echo $file && sed -i "s/ *$//" $file;
done
```

Only removes trailing *spaces*, if you truly want to remove trailing *whitespace* you will also need to remove tabs and other more esoteric white-space characters.

Try:
```bash
for file in $(find . -type f -name \*.php | grep -v "^./vendor" | grep -v "^./lib/Doctrine/compiledEntities/"); do
    echo $file && sed -i "s/\s*$//" $file;
done
```

GRyall:
I have now rerun that and squashed it into the commit. I had thought I would be cautious and that lines were less likely to end with tabs than spaces (there were 290 lines). That said, the commit message was still wrong.

GRyall:
Following the suggestion of @davidmeredith,  I have redone the tab replacement in the html files and used two instead 4 spaces.

GRyall:
Leading white space commits have now been split into pull request #29 . When that is merged I will look again at trailing space.

PR #28, opened 2016-06-16
Leading whitespace

GRyall:
Split from #27
- Replaces leading tabs with 4 spaces each on 8910 lines of 211 .php files
- Replaces leading tabs with 2 spaces each on 170 lines of 5 html files


GRyall:
Should have been a request on dev

PR #41, opened 2016-09-28
Allow ':' in HOST_DN field

GRyall:
Allows for the use of colons in DNs


GRyall:
Ultimately it was decided this was not a good solution and that colons shouldn't appear in Host DNs.

PR #79, opened 2017-02-07
Provide a configuration for travis to run unit tests

jrha:
Add a self-contained bootstrap configuration and setup script under `.travis/` that can be run in a CI environment

Currently runs tests using PHP 5.3 against a sqlite database backend, but this can be extended to a larger build matrix in `.travis.yml` by:
* Adding more versions of PHP to the `php` list.
* Adding more DB entries to the `env` list with matching bootstrap configs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: php
+before_script: .travis/setup
+script: vendor/bin/phpunit tests/DoctrineTestSuite1.php
+php:
+  - '5.3'
+env:
+  - DB=sqlite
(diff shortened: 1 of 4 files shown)


jrha:
In order to get the tests running, the owner of the @gocdb will need to sign in to https://travis-ci.org/ and activate the repository.

Note that some tests currently fail, this appears to be related to sqlite, but I haven't proved this exhaustively.

jrha:
Oh and for an example of what travis will output, you can look at my testing at https://travis-ci.org/jrha/gocdb

PR #81, opened 2017-02-08
Correct capitalisation of table names in unit tests

jrha:
As MySQL tables are files on disk, MySQL follows the  case-sensitivity of the Operating System,
so tests using mangled capitalisation fail.

These may have been missed before as Oracle and MySQL on Windows are case-insensitive.

This corrects the capitalisation to match that generated by Doctrine.


GRyall:
(That's oracle in general that is case insensitive)

PR #92, opened 2017-06-14
Change from assurance LoA attribute to entitlement

davidmeredith:
Updates needed to support EGI CheckIn service which has changed from providing assurance attribute to entitlement attribute.


GRyall:
This already appears to be in place on the production GOCDB server, as a manual change. I am merging this to allow us a quick path to new updates.

PR #95, opened 2017-10-27
Add Precise MySQL Travis builds

tofu-rocketry:
- Add Precise MySQL builds to Travis config which should enable PHP 5.3 tests to work.
- Allow Trusty PHP 5.3 build to fail as this combination is no longer supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+# N.B. Ubuntu Precise available on Travis until March 2018
+dist: precise
+sudo: required
 language: php
 before_install:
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
Bah, seems like you have to use `include` in the matrix to do multi-dist builds. 😠

tofu-rocketry:
N.B. Ubuntu Precise will only be available on Travis until March 2018.

Builds x.4 and x.5 should be the Trusty ones. Rest will be Precise.

tofu-rocketry:
Had to give the last build a boot up the bum.

PR #98, opened 2017-11-06
Correct merge into master instead of Dev

GRyall:
Pull request #97 should have been merged into dev. Given we are testing at the moment anyway, it is harmless to leave it in master, so we'll merge master back into dev.


tofu-rocketry:
Fast-forward merge performed instead.

PR #105, opened 2018-03-02
travis: Use built-in PHP syntax checker on modified PHP files

jrha:
The test suite doesn't cover everything, so this seems like a low-effort way to get some extra coverage for free.

### DIFF
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+if [[ $TRAVIS ]]; then
+    cd "$TRAVIS_BUILD_DIR" || exit 2
+
+    # Check modified PHP files with PHP's internal syntax checker
+    git diff --name-only HEAD^ | grep '\.php$' | xargs -r -n 1 php -l || exit 1
+
+    # Run test suite
+    vendor/bin/phpunit tests/DoctrineTestSuite1.php
+else
+    echo 'ABORTED: NOT RUNNING ON TRAVIS'
+    exit 2
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: php
 before_install:
     - mysql -e 'CREATE DATABASE doctrine;'
 before_script: .travis/setup
-script: vendor/bin/phpunit tests/DoctrineTestSuite1.php
+script: .travis/run_tests.sh
 services:
     - mysql
 php:
--- a/htdocs/PI/index.php
+++ b/htdocs/PI/index.php
@@ -74,7 +74,7 @@ class PIRequest {
     private $params = array();
     private $dn = null;
     private $baseUrl;
-    private $baseApiUrl; 
+    private $baseApiUrl;
 
     // params used to set the default behaviour of all paging queries,
     // these vals can be overidden per query if needed.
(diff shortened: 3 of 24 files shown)


jrha:
This shouldn't cause any new failures, in a moment I will modify some PHP files to check this works properly.

jrha:
Tests with modified files passed fine, Travis is being a bit slow today.

.travis/run_tests.sh
```diff
@@ -0,0 +1,14 @@
...
+    cd "$TRAVIS_BUILD_DIR" || exit 2
+
+    # Check modified PHP files with PHP's internal syntax checker
+    git diff --name-only HEAD^ | grep '\.php$' | xargs -r -n 1 php -l || exit 1
+
+    # Run test suite
+    vendor/bin/phpunit tests/DoctrineTestSuite1.php
+else
+    echo 'ABORTED: NOT RUNNING ON TRAVIS'
+    exit 2
```
tofu-rocketry: Why an exit code of 2?

.travis/run_tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
```
tofu-rocketry: As this file only runs tests for Travis, should it have "travis" in the filename?

.travis/run_tests.sh
```diff
@@ -0,0 +1,14 @@
...
+if [[ $TRAVIS ]]; then
+    cd "$TRAVIS_BUILD_DIR" || exit 2
+
+    # Check modified PHP files with PHP's internal syntax checker
+    git diff --name-only HEAD^ | grep '\.php$' | xargs -r -n 1 php -l || exit 1
```
tofu-rocketry: So this runs on the _whole_ of any file that has been changed, right?

.travis/run_tests.sh
GRyall: it is in a directory called '.travis'

.travis/run_tests.sh
jrha: Because `1` would indicate test failure, `2` is a standard exit code for "usage error".

.travis/run_tests.sh
jrha: Yup.

.travis/run_tests.sh
tofu-rocketry: Doesn't Travis treat any non-zero exit code in the `script` section as a fail?

.travis/run_tests.sh
tofu-rocketry: I had not spotted that. :eyeglasses:

.travis/run_tests.sh
tofu-rocketry: Is that for efficiency, or are there some files that would currently fail?

jrha: Not for efficiency, but it seems cruel to report errors on files that have not been changed.

.travis/run_tests.sh
jrha: Yes indeed, but this is to catch the case of a user running the script manually, Travis wouldn't get here.

.travis/run_tests.sh
tofu-rocketry: :sparkles: Fairy Nuff. :sparkles:

PR #106, opened 2018-03-13
Ensure reserved characters are escaped in links

### LINKED ISSUE #103: Ampersands in link urls should be encoded
Many (maybe all) link hrefs do not encode ampersands in URLs, for example the following:
```html
<a href="index.php?Page_Type=Static_HTML&Page=Help_And_Contact">
```
should be:
```html
<a href="index.php?Page_Type=Static_HTML&amp;Page=Help_And_Contact">
```
This is necessary because `&` is a an escape character for HTML entities and is therefore a reserved character, see https://tools.ietf.org/html/rfc3986#section-2.2.

jrha:
Fixes #103, but is not restricted to just ampersands.

### DIFF
--- a/htdocs/web_portal/components/Draw_Components/menu.php
+++ b/htdocs/web_portal/components/Draw_Components/menu.php
@@ -101,7 +101,7 @@ function add_menu_item($menu_item)
     if ($show == "all" or ($show=="write_enabled" and (!$portalIsReadOnly or $userisadmin)) or (($show == "admin") and ($userisadmin))){
         $html .=    	"<li class=\"Menu_Item\">".
                         "<a href=\"".htmlspecialchars($link)."\"><span class=\"menu_link\">".
-        $name."</span></a></li>";
+        htmlspecialchars($name)."</span></a></li>";
     }
     return $html;
 }
--- a/htdocs/web_portal/components/Draw_Components/menu.php
+++ b/htdocs/web_portal/components/Draw_Components/menu.php
@@ -100,7 +100,7 @@ function add_menu_item($menu_item)
     }
     if ($show == "all" or ($show=="write_enabled" and (!$portalIsReadOnly or $userisadmin)) or (($show == "admin") and ($userisadmin))){
         $html .=    	"<li class=\"Menu_Item\">".
-                        "<a href=\"".$link."\"><span class=\"menu_link\">".
+                        "<a href=\"".htmlspecialchars($link)."\"><span class=\"menu_link\">".
         $name."</span></a></li>";
     }
     return $html;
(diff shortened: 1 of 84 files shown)


htdocs/web_portal/views/scope_help.php
```diff
@@ -44,26 +44,26 @@
               from all the child Services - a Service can't have a reserved tag that
               is not supported by its parent Site.</li>
             <li>For NGIs: If a Reserved tag is removed from an NGI, then the same tag is NOT
-              removed from all the child Sites - this is intentionally different from the Site->Service relationship.</li>
+              removed from all the child Sites - this is intentionally different from the Site&rarr;Service relationship.</li>
```
tofu-rocketry: Is this just a nicety or does the angle bracket cause problems?

htdocs/web_portal/components/Draw_Components/menu.php
```diff
@@ -100,10 +100,10 @@ function add_menu_item($menu_item)
     }
     if ($show == "all" or ($show=="write_enabled" and (!$portalIsReadOnly or $userisadmin)) or (($show == "admin") and ($userisadmin))){
         $html .=    	"<li class=\"Menu_Item\">".
-                        "<a href=\"".$link."\"><span class=\"menu_link\">".
+                        "<a href=\"".htmlspecialchars($link)."\"><span class=\"menu_link\">".
```
tofu-rocketry: How's this work?

GRyall: http://php.net/manual/en/function.htmlspecialchars.php

htdocs/web_portal/views/scope_help.php
jrha: Both.
1. This is rich text, in this case a Unicode right arrow conveys the intended meaning better.
2. The `>` could be interpreted as the end of a tag, especially if there was a stray `<` further up.

htdocs/web_portal/components/Draw_Components/menu.php
tofu-rocketry: Ah, ok. I wasn't sure if it was a PHP thing or your own function that was somewhere else.

tofu-rocketry:
Any reason this can't be merged?

tofu-rocketry:
Do you mean code review or testing each change?

jrha:
I really hope you don't mean testing each change.

htdocs/web_portal/components/Draw_Components/menu.php
```diff
@@ -100,10 +100,10 @@ function add_menu_item($menu_item)
...
         $name."</span></a></li>";
```
jrha: I've just noticed that the menu item names also contain ampersands and so also need escaping, I will add another commit.

PR #111, opened 2018-05-10
Fix host dn regexp

### LINKED ISSUE #110: Make DN regexes comply with OGF standard
As reported [on GGUS here](https://ggus.eu/index.php?mode=ticket_info&ticket_id=134788).

The [regex for host DNs](https://github.com/GOCDB/gocdb/blob/dev/config/gocdb_schema.xml#L143) is overly picky when compared to [the OGF standard (section 4.3)](https://www.ogf.org/documents/GFD.225.pdf). In this specific case of the submitted GGUS ticket we're disallowing parentheses even though they are valid, but we should also check generally that we match the standard.

This may also apply to the other ...`_DN` regexes.

ineilson:
Resolves #110 

Copied existing regexp for user certificate dn (which is commented out but includes bracket characters) to be used for the host dn.

### DIFF
--- a/config/gocdb_schema.xml
+++ b/config/gocdb_schema.xml
@@ -140,7 +140,7 @@
 	<field>
 	    <fname>HOST_DN</fname>
 	    <length>255</length>
-	    <regex>/^(\/[a-zA-Z]+=[a-zA-Z0-9\-\_\s\.@,'\/\)\(]+)+$/</regex>
+	    <regex>/^(\/[a-zA-Z]+=[a-zA-Z0-9\-\_\s\.@,'\/\)\(]+)*$/</regex>
 	</field>
 	<field>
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
There seem to be two other differences with the regex:
- It removes the duplicated literal `/`
- It now matches one or more, rather than zero or more, or the main group.

The swapping of the `a-z` and `A-Z` probably doesn't make a difference.

GRyall:
What is the logic for the switch to one or more from zero or more?  I am concerned that this may prevent empty string being the value. I.e. if no DN is specified, does the vailidation still pass for this DN?

ineilson:
Yes, of course, without checking the program logic, it was a too naive assumption that an empty string is an invalid DN in context. I'll revisit.

PR #113, opened 2018-07-02
Updates to installation doc

ineilson:
Include misc. changes to improve installation documentation.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,10 +14,10 @@ This file is best viewed using a browser-plugin for markdown `.md` files.
   * v5.3.3 (newer versions should be fine, but are untested)
   * If using Oracle: PHP oci8 extension (needs to be compiled using the Oracle Instant client v10 or higher
     downloadable from Oracle website, see "Compiling OCI8" section below).
-  * libxml2 and DOM support for PHP (Note: On RHEL, PHP requires the PHP XML RPM to be installed for this component to function).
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -392,9 +402,6 @@ This directory object defines the directory where the .dmp file is loaded from.
   ```
   sqlplus system
   SQL> create or replace DIRECTORY dmpdir AS '<Directroy path>';
-  SQL> grant read,write on directory dmpdir to <user>;
```
GRyall: Reason for deltion?

INSTALL.md
```diff
@@ -117,18 +117,26 @@ rpm -i oracle-instanclient*
...
 
+Install the oci8 module using pecl:
+
+```bash
+pecl install oci8-2.0.10
+```
 This will download and compile the module, and place it in your php extension dir.
 
-Add the ```extension=oci8.so``` line to your php.ini. Confirm it is working with ```php -i | grep -i oci8```
+Add the ```extension=oci8.so``` line to your php.ini or create a configuration file:
```
GRyall: Actually - I would suggest there is no or. Users _should_ be creating a seperate  ini configuration file.

tofu-rocketry: Did you not want this changed before merging?

INSTALL.md
tofu-rocketry: Did you get an answer to this?

PR #121, opened 2018-07-13
Replaced all tabs with 4 spaces in the CSS for both the landing page and portal.

CoolCoder2019:
Merge in after #120 and rebase so that this PR only changes white space.


GRyall:
There is also a colour fix in this for a hex-code which should have it's own commit. But life's too short.

PR #127, opened 2018-09-10
mariadb stores table Ngis_Scopes

nikoloutsa:
After updating to the latest version i got an error for the table `NGIs_Scopes` as non existing, although the table is created in the DB with a lower case name `Ngis_Scopes`
```
An exception occurred while executing 'SELECT o0_.id AS id0, n1_.name AS name1, n1_.email AS email2, n1_.rodEmail AS rodEmail3, n1_.helpdeskEmail AS helpdeskEmail4, n1_.securityEmail AS securityEmail5, n1_.description AS description6, n1_.ggus_Su AS ggus_Su7, n1_.creationDate AS creationDate8, s2_.id AS id9, s2_.name AS name10, s2_.description AS description11, o0_.discr AS discr12 FROM NGIs n1_ INNER JOIN OwnedEntities o0_ ON n1_.id = o0_.id LEFT JOIN NGIs_Scopes n3_ ON n1_.id = n3_.ngi_Id LEFT JOIN Scopes s2_ ON s2_.id = n3_.scope_Id ORDER BY n1_.name ASC': SQLSTATE[42S02]: Base table or view not found: 1146 Table 'GOCDB5.NGIs_Scopes' doesn't exist
```

### DIFF
--- a/lib/Doctrine/entities/NGI.php
+++ b/lib/Doctrine/entities/NGI.php
@@ -66,7 +66,7 @@ class NGI extends OwnedEntity implements IScopedEntity {
      * Unidirectional - Scope tags associated with the NGI
      *
      * @ManyToMany(targetEntity="Scope")
-     * @JoinTable(name="NGIs_Scopes",
+     * @JoinTable(name="Ngis_Scopes",
      *      joinColumns={@JoinColumn(name="ngi_Id", referencedColumnName="id")},
      *      inverseJoinColumns={@JoinColumn(name="scope_Id", referencedColumnName="id")}
      *      )


tofu-rocketry:
What environment are you using? It seems like your database is doing something overly clever with capitalisation.

I don't think this change is the correct one as the table should be `NGIs_Scopes` (as per the unit tests which are now failing). There may be something that needs setting to enforce the capitalisation as originally intended when creating the table.

nikoloutsa:
I am using mariadb , i just have the defaults for capitalization [lower_case_table_names set to 0 the default on Unix-based systems](https://mariadb.com/kb/en/library/server-system-variables/#lower_case_table_names)
```
mysql --version
mysql  Ver 15.1 Distrib 5.5.56-MariaDB, for Linux (x86_64) using readline 5.1
```
```
MariaDB [GOCDB5]> show tables;
+-----------------------------+
| Tables_in_GOCDB5            |
+-----------------------------+
| NGIs                        |
| Ngis_Scopes                 |
```

> I don't think this change is the correct one as the table should be NGIs_Scopes (as per the unit tests which are now failing).

Yes this should fail the unit tests have a hard coded value 'NGIs_Scopes' , sorry for not changing that.

> There may be something that needs setting to enforce the capitalisation as originally intended when creating the table.

Where exactly is the table created? Doctrine successfully updates with lowercase, here is sql dump: from 5.7-->5.7.2 

```
$ ../../vendor/bin/doctrine orm:schema-tool:update --dump-sql
ALTER TABLE Ngis_Scopes DROP FOREIGN KEY FK_69BAE37BFDAF7D93;
ALTER TABLE Ngis_Scopes DROP FOREIGN KEY FK_69BAE37BE284E4DD;
DROP INDEX idx_a4166208e284e4dd ON Ngis_Scopes;
CREATE INDEX IDX_69BAE37BE284E4DD ON Ngis_Scopes (ngi_Id);
DROP INDEX idx_a4166208fdaf7d93 ON Ngis_Scopes;
CREATE INDEX IDX_69BAE37BFDAF7D93 ON Ngis_Scopes (scope_Id);
ALTER TABLE Ngis_Scopes ADD CONSTRAINT FK_69BAE37BFDAF7D93 FOREIGN KEY (scope_Id) REFERENCES Scopes (id);
ALTER TABLE Ngis_Scopes ADD CONSTRAINT FK_69BAE37BE284E4DD FOREIGN KEY (ngi_Id) REFERENCES NGIs (id);
ALTER TABLE Sites DROP notify;
ALTER TABLE Services DROP notify;
```

nikoloutsa:
I think my problem is from the commit [Correct capitalisation of table names](https://github.com/GOCDB/gocdb/commit/9d18a12069a05def1edb06fb07cc87e73e17b9d1#diff-041063f4619ec0782e4cce1fb4489554)

You have altered the table names to 'NGIs_Scopes' but I have already created the database with version gocdb5.7:
```
$ ../../vendor/bin/doctrine orm:schema-tool:create --dump-sql | grep NGIs_Scopes
CREATE TABLE NGIs_Scopes (ngi_Id INT NOT NULL, scope_Id INT NOT NULL, INDEX IDX_A4166208E284E4DD (ngi_Id), INDEX IDX_A4166208FDAF7D93 (scope_Id), PRIMARY KEY(ngi_Id, scope_Id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB;
```
the same command on gocdb5.7.2
```
$ ../../vendor/bin/doctrine orm:schema-tool:create --dump-sql | grep NGIs_Scopes
CREATE TABLE NGIs_Scopes (ngi_Id INT NOT NULL, scope_Id INT NOT NULL, INDEX IDX_A4166208E284E4DD (ngi_Id), INDEX IDX_A4166208FDAF7D93 (scope_Id), PRIMARY KEY(ngi_Id, scope_Id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB;
```

jrha:
This was a bug in previous versions of the code-base, the table really should be called `NGIs_Scopes`, if you have upgraded from a previous version you will have to rename the existing table. Apologies for not having drawn attention to that in the release notes.

FYI the bug wasn't picked up sooner because the database behind the production GOCDB instance is not case-sensitive and for a long time the unit tests were being run only on Windows (again not case sensitive).

GRyall:
As @jrha says, this fixes a bug in the first place, so I am reluctant to revert. That said, we should have made the release notes clearer - at the time I hadn't realised anyone else was running GOCDB. 

The fix from your perspective would be to run an sql command against the table to rename it . I beleive something like this:
` ALTER TABLE Ngis_Scopes RENAME TO NGIs_Scopes`

Thankyour for raising this as a pull request - in other circumstances we would have happily merged it.

PR #139, opened 2019-01-31
Exclude deleted files from PHP lint checks

jrha:
Otherwise tests will fail, which will make us sad.

### DIFF
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -4,7 +4,7 @@ if [[ $TRAVIS ]]; then
     cd "$TRAVIS_BUILD_DIR" || exit 2
 
     # Check modified PHP files with PHP's internal syntax checker
-    git diff --name-only HEAD^ | grep '\.php$' | xargs -r -n 1 php -l || exit 1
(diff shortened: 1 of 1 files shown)


.travis/run_tests.sh
```diff
@@ -4,7 +4,7 @@ if [[ $TRAVIS ]]; then
     cd "$TRAVIS_BUILD_DIR" || exit 2
 
     # Check modified PHP files with PHP's internal syntax checker
-    git diff --name-only HEAD^ | grep '\.php$' | xargs -r -n 1 php -l || exit 1
+    git diff --name-only --diff-filter=ACMRTUXB HEAD^ | grep '\.php$' | xargs -r -n 1 php -l || exit 1
```
tofu-rocketry: Ah, so it was checking _all_ modified files, including deleted ones? This change filters for everything other than Deleted (`D`)?

tofu-rocketry:
[`git-diff --diff-filter` Documentation](https://git-scm.com/docs/git-diff#git-diff---diff-filterACDMRTUXB82308203)

PR #143, opened 2019-03-20
Increment version number to 5.7.3

gregcorbett:
in preparation for a merge of `dev` to `master` and deployment on gocdb-test


gregcorbett:
this code is running [here](https://host-172-16-113-91.nubes.stfc.ac.uk/portal/).

gregcorbett:
Seems to be based on recursively grep-ing for "5.7.2" and based on [this](https://github.com/GOCDB/gocdb/commit/e5782c7dc76811f2d40c807a417ab71cb76ec979).

tofu-rocketry:
Fair enough. Just surprised compared to all the places we have it in APEL.

My other comment is that the body font is too small. It's 9pt, but 10pt is really as small as you should go.

gregcorbett:
> My other comment is that the body font is too small. It's 9pt, but 10pt is really as small as you should go.

Yeah, fair enough. See #144.

GRyall:
Just to confirm that the version number is indeed only in one place

PR #146, opened 2019-04-05
Allow multiemail input for ServiceGroup EMAIL

nikoloutsa:
For our use-case it would make sense to allow more than one emails for ServiceGroup contact , since it's not a person it's a group.

- test1@mail.com valid
- test1@mail.com;test2@mail.com valid
- test1@mail.com; invalid
- test1@mail.com;test2@mail.com; invalid

	modified:   config/gocdb_schema.xml
	modified:   htdocs/web_portal/views/service_group/edit_service_group.php


htdocs/web_portal/views/service_group/edit_service_group.php
```diff
@@ -9,7 +9,7 @@
         <input type="text" value="<?php xecho($sg->getName()) ?>" name="name" class="input_input_text">
         <span class="input_name">Description</span>
         <input type="text" value="<?php xecho($sg->getDescription()) ?>" name="description" class="input_input_text">
-        <span class="input_name">Contact E-Mail *<span class="input_syntax" >(valid email format)</span></span>
+        <span class="input_name">Contact E-Mail *<span class="input_syntax" >(valid email format semicolong separated)</span></span>
```
jrha:
```suggestion
        <span class="input_name">Contact E-Mail *<span class="input_syntax" >(valid email format semicolon separated)</span></span>
```

config/gocdb_schema.xml
```diff
@@ -380,7 +380,7 @@
 	<field>
 	    <fname>EMAIL</fname>
 	    <length>255</length>
-	    <regex>/^([0-9a-zA-Z]+[-._])*[0-9a-zA-Z]+@([-0-9a-zA-Z]+[.])+[a-zA-Z]{2,6}$/</regex>
+	    <regex>/^(([0-9a-zA-Z]+[-._])*[0-9a-zA-Z]+?@([-0-9a-zA-Z]+[.])+[a-zA-Z]{2,6})(;(([0-9a-zA-Z]+[-._])*[0-9a-zA-Z]+?@([-0-9a-zA-Z]+[.])+[a-zA-Z]{2,6}))?$/</regex>
```
jrha: This only matches one OR two email addresses, if you want to be able to match any number then you probably want:
```suggestion
	    <regex>/^(([0-9a-zA-Z]+[-._])*[0-9a-zA-Z]+?@([-0-9a-zA-Z]+[.])+[a-zA-Z]{2,6})(;(([0-9a-zA-Z]+[-._])*[0-9a-zA-Z]+?@([-0-9a-zA-Z]+[.])+[a-zA-Z]{2,6}))*$/</regex>
```

nikoloutsa: indeed , thanks a lot !

gregcorbett: What is the benefit of adding the `?` before the `@` symbols?

gregcorbett:
>     * [test1@mail.com](mailto:test1@mail.com);[test2@mail.com](mailto:test2@mail.com) valid
>     * [test1@mail.com](mailto:test1@mail.com);[test2@mail.com](mailto:test2@mail.com) invalid

What is the difference between these two cases? To me, they look the same.

config/gocdb_schema.xml
jrha: Well it makes the `+` operator lazy, which I can't obviously see a benefit of in this case.

jrha:
Slightly off-topic but maybe something to consider for @gregcorbett as a later improvement: The base regex may be too restrictive as it definitely does not allow for Unicode characters in addresses, which may or may not be allowed anyway.

It might also be worth seeing whether PHP's built-in regex filter would be good enough:
```php
filter_var($email, FILTER_SANITIZE_EMAIL);
```

The [HTML5 spec says](https://html.spec.whatwg.org/#e-mail-state-(type=email)):
```regex
/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
````

So… :man_shrugging:

File: config/gocdb_schema.xml (hunk 2 above)
Comment (tofu-rocketry) [reply config/gocdb_schema.xml]: Shall we remove it?

**Comment by gregcorbett** [comment]
> 
> 
> Slightly off-topic but maybe something to consider for @gregcorbett as a later improvement: The base regex may be too restrictive as it definitely does not allow for Unicode characters in addresses, which may or may not be allowed anyway.
> 
> It might also be worth seeing whether PHP's built-in regex filter would be good enough:
> 
> ```
> filter_var($email, FILTER_SANITIZE_EMAIL);
> ```
> 
> The [HTML5 spec says](https://html.spec.whatwg.org/#e-mail-state-(type=email)):
> 
> ```regexp
> /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
> ```
> 
> So… 🤷‍♂️

Feel free to open an issue :)

**Comment by tofu-rocketry** [comment]
@jrha opened an issue (#147). Do we just need to sort the lazy matching out?

**Comment by gregcorbett** [comment]
> @jrha opened an issue (#147). Do we just need to sort the lazy matching out?

And what the difference between these [two test cases](https://github.com/GOCDB/gocdb/pull/146#issuecomment-480752408) is suppose to be. @nikoloutsa?.

**Comment by nikoloutsa** [comment]
the difference is not to all a semicolon at the end of the mail field. 

It looked cleaner to me when someone inputs just a single email, to trim out the last semicolon as it was previously without my addition.

**Comment by tofu-rocketry** [comment]
> the difference is not to allow a semicolon at the end of the mail field.

@nikoloutsa, There doesn't seem to be a semicolon at the end of the last test case though. I can see that there is a semicolon at the end of the invalid case that only has one email address, so was that just a mistype for the invalid two address case?

**Comment by GRyall** [comment]
As discussed above I have removed the lazy matching. Following in person discussions with the devlopment team I have also reworked the regex to allow trailing semicolons. @gregcorbett this is ready to be merged!

File: htdocs/web_portal/views/service_group/edit_service_group.php [hunk 3]
```diff
@@ -9,7 +9,7 @@
         <input type="text" value="<?php xecho($sg->getName()) ?>" name="name" class="input_input_text">
         <span class="input_name">Description</span>
         <input type="text" value="<?php xecho($sg->getDescription()) ?>" name="description" class="input_input_text">
-        <span class="input_name">Contact E-Mail *<span class="input_syntax" >(valid email format)</span></span>
+        <span class="input_name">Contact E-Mail *<span class="input_syntax" >(valid email format semicolon separated)</span></span>
```
gregcorbett:
```suggestion
        <span class="input_name">Contact E-Mail *<span class="input_syntax" >(valid email format, semicolon separated without spaces)</span></span>
```

PR #155, opened 2019-06-12
 Allow "org.squid-cache.Squid" to be production and not monitored and improve handling of these cases

gregcorbett:
As per https://ggus.eu/index.php?mode=ticket_info&ticket_id=139924

- Change the message on "production =/> monitored". Rather than list an potentially ever growing list of ServiceTypes, just say that the 'production => monitored' rule applies to the ServiceType in question. When we move to having this rule as a property of the ServiceType, we probably wouldn't want to enumerate the exceptions anyway.
- Replace three `if` statements with an `in_array` call. Atleast for now, once this is a property of the service type this may change again. This seems cleaner than a potentially ever growing number of `if` statements.
- Move 'production => monitored' rule into function

I'll open an issue stating that the 'production => monitored' rule should be a property of the ServiceType in the database and not hardcoded. But I want to be able to throw up databases in a quattorized fashion before tackling that.

### DIFF
--- a/lib/Gocdb_Services/ServiceService.php
+++ b/lib/Gocdb_Services/ServiceService.php
@@ -511,7 +511,7 @@ class ServiceService extends AbstractEntityService {
         // validate production/monitored combination
         if ($st != 'VOMS' && $st != 'emi.ARGUS' && $st != 'org.squid-cache.Squid') {
             if ($newValues ['PRODUCTION_LEVEL'] == "Y" && $newValues ['IS_MONITORED'] != "Y") {
-                throw new \Exception ( "If Production flat is set to True, Monitored flag must also be True (except for VOMS, emi.ARGUS and org.squid-cache.Squid)" );
+                throw new \Exception ( "For the '".$st."' service type, if the Production flag is set to True, the Monitored flag must also be True." );
             }
         }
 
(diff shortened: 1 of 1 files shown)


gregcorbett:
> I'll open an issue stating that the 'production => monitored' rule should be a property of the ServiceType in the database and not hardcoded. But I want to be able to throw up databases in a quattorized fashion before tackling that.

Issue #156 created.

lib/Gocdb_Services/ServiceService.php
```diff
@@ -509,11 +538,7 @@ public function editService(\Service $se, $newValues, \User $user = null) {
         $this->validateEndpointUrl ( $newValues ['endpointUrl'] );
         $this->uniqueCheck ( $newValues ['SE'] ['HOSTNAME'], $st, $se->getParentSite () );
         // validate production/monitored combination
-        if ($st != 'VOMS' && $st != 'emi.ARGUS') {
-            if ($newValues ['PRODUCTION_LEVEL'] == "Y" && $newValues ['IS_MONITORED'] != "Y") {
-                throw new \Exception ( "If Production flat is set to True, Monitored flag must also be True (except for VOMS and emi.ARGUS)" );
```
tofu-rocketry: Good to get that "flat" replaced!

lib/Gocdb_Services/ServiceService.php
```diff
@@ -459,6 +459,35 @@ public function getServiceTypes() {
...
+     * Validates the the 'production => monitored' rule for the user inputted 
+     * service data.
+
+     * @param array $serviceValues
+     * @throws \Exception If the serviceValues production/monitored combination
+     * is invalid. The \Exception's message will contain a human readable error
+     * message.
+     * @return null
+     */
+    private function validateProductionMonitoredCombination($serviceValues) {
```
tofu-rocketry: Good to have this factored out. I guess this'll make using a config value easier.

PR #157, opened 2019-07-03
Refactor to make service URLs less exceptional and minor other things

GRyall:
* For legacy reasons the URLS of services have been treated as a seperate
entity for the purposes of validation. This no longer makes sense, so I
have updated the codebase to reflect that url is just another property
of a service.

PR #161, opened 2019-07-03
Validate types of entity fields

GRyall:
Depends on https://github.com/GOCDB/gocdb/pull/157 

* adds entries in the GOCDB schema for boolean fields of services and endpoints
* Introduces the ability to validate boolean and string properties of
entities defined in the GOCDB schema. This change is designed to not
interfere with existing behaviour of the validation tool. 
* Boolean fields in the schema are marked as such.
* Make use of new validation for SEs in web portal

### DIFF
--- a/config/gocdb_schema.xml
+++ b/config/gocdb_schema.xml
@@ -169,6 +169,15 @@
       <!-- Support for all URI chars (unreserved, gen-delims, sub-delims and percent) RFC 3986 -->
 	    <regex>/^[0-9a-zA-Z:\/\.\&amp;\?\_\$\+\!\*\'\(\)\,\-#\[\]@;=~%]*$/</regex>
   </field>
+  <field>
+      <fname>PRODUCTION</fname>
+  </field>
+  <field>
+      <fname>BETA</fname>
+  </field>
+  <field>
+      <fname>MONITORED</fname>
+  </field>
     </entity>
     <!-- ==========================================================  -->
     <entity>
@@ -354,6 +363,9 @@
         <fname>EMAIL</fname>
         <length>255</length>
         <regex>/^(([0-9a-zA-Z]+[-._])*[0-9a-zA-Z]+@([-0-9a-zA-Z]+[.])+[a-zA-Z]{2,6}[,;]?)*$/</regex>
+    </field>
+    <field>
+        <fname>MONITORED</fname>
     </field>
     </entity>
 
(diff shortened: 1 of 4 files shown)


lib/Gocdb_Services/Validate.php
```diff
@@ -36,6 +36,22 @@ public function validate($Object_Name, $Field_Name, $Field_Value)
...
+        //Check the Type of the field value, where this is defined
+        $type = $this->Get_Field_value($Object_Name, $Field_Name, 'ftype');
+        if (count($type) != 0) {
+          if (strtolower($type) == "string") {
+            if (!is_string($Field_Value)) {
+                throw new \Exception($Field_Name . " must be a string.");
+            }
+          } elseif (strtolower($type) == "boolean") {
+            if (!is_bool($Field_Value)) {
+                throw new \Exception($Field_Name . " must be a boolean.");
```
tofu-rocketry: Should we be more specific about the accepted values? Different languages use different things as bools.

GRyall: I am sorry, I don't understand your question.  Which bit should be more specific?

tofu-rocketry: Sorry. In the exception message.

GRyall: The exception could come about because a developer has failed to ensure the correct type of thing has made it to this function, in which case the developer knows it is Php. It could occur due to validation in the write API, in which case the nature of what a boolean is dependent on the input language being used (currently only JSON is supported). In the case of the write API, the exception is caught and a link to the documentation is appended. This documentation makes clear what is expected for booleans.

tofu-rocketry: Ok, fair enough.

lib/Gocdb_Services/Validate.php
```diff
@@ -36,6 +36,22 @@ public function validate($Object_Name, $Field_Name, $Field_Value)
         //Check the length of the field value, if too long, throws exception
         $this->checkFieldLength($Object_Name, $Field_Name, $Field_Value);
 
+        //Check the Type of the field value, where this is defined
+        $type = $this->Get_Field_value($Object_Name, $Field_Name, 'ftype');
+        if (count($type) != 0) {
```
gregcorbett: Does `Get_Field_value` returns an array?

GRyall: I haven't checked, but I suspect it returns an array with one element which is the value of that field, or an empty array if it couldn't find anything. This snippet of code was copied from elsewhere in the validation code.

config/gocdb_schema.xml
```diff
@@ -163,6 +163,25 @@
 	    <length>255</length>
 	    <regex>/^(([0-9a-zA-Z]+[-._])*[0-9a-zA-Z]+@([-0-9a-zA-Z]+[.])+[a-zA-Z]{2,6}[,;]?)+$/</regex>
 	</field>
+  <field>
```
gregcorbett: This block seems differently indented to [this](https://github.com/GOCDB/gocdb/pull/161/files#diff-950b814119a26dd29c62aac47af704adR371) block (2 spaces versus 4 spaces) in the same file.

It will be easier if we keep the new indenting consistent with the old in the file.

PR #173, opened 2019-08-14
Trim all free text entry form fields

### LINKED ISSUE #82: Whitespace trimming on email (all?) fields
It would improve the user experience if we were to trim leading and trailing white space from submitted email addresses in forms (this has been pointed out to me for the add/edit service ones in particular). The error message when such white space is present can be unclear.

tofu-rocketry:
Resolves #82.

- Trim all free text entry form fields so that extraneous whitespace
doesn't make it either to the validation stage for some fields (leading
to an unhelpful error message) or into the database (leading to problems
with some services that use data from GOCDB).


tofu-rocketry:
I need help getting this onto a test instance of some sort.

PR #179, opened 2019-09-12
Fix typos in comments and a small bug in Write API

GRyall:
Fixes a couple of comment typos and removes a bug in the write API


tofu-rocketry:
> The switch variable was a statement that evaluated to true (it probably used to be an if statement and was replaced with a switch at some point). Luckily, for reasons, the correct case was still called (or not), so it was fine. had there been multiple switch cases, only the first would have been called.

Ah, got it, thanks.

PR #185, opened 2019-10-24
Update STFC logos to reflect new branding

GRyall:
This update replaces the old STFC logo with the new version. There is a cropped version for the portal, which is likely technically not in line with brand guidelines. However, the old cropped logo it replaces was not in line with the old guidelines either...

image

image


tofu-rocketry:
What using about the "square lock-up", as they call it, for the small logos? Is the text in that just too small?

GRyall:
> 
> 
> What using about the "square lock-up", as they call it, for the small logos? Is the text in that just too small?

It's not available for download from the page with all the logos. but here is a mock up:
image

tofu-rocketry:
That is quite tiny. 🤔

jrha:
Hmm, this is going to keep bugging me about the new logo, but the little tick at the top bothers me, as it shrinks the box and breaks the horizontal lines in the logo box, it might be too messy to fix, but could it be slightly larger?

Mockup:
image

jrha:
Also... maybe now is the time to use vector based logos?

ineilson:
> 
> 
> Hmm, this is going to keep bugging me about the new logo, but the little tick at the top bothers me, as it shrinks the box and breaks the horizontal lines in the logo box, it might be too messy to fix, but could it be slightly larger?
> 
> Mockup:
> image

Was just thinking exactly the same. That or move the UKRI box away from the EU one which dominates.

GRyall:
I certainly agree about the need to use vector based logos. I am a little worried about getting different height logos to align correctly in a simple way.

tofu-rocketry:
Shall we not worry about this for now for the purposes of this PR?

jrha:
Pfft fine, I'll allow it.

tofu-rocketry:
> Pfft fine, I'll allow it.

"Perfect is the enemy of good." ~ James Adams after Voltaire

jrha:
I said I'd allow it…

PR #184, opened 2019-10-24
Add support for EGI Dev IdP out of the box

gregcorbett:
I have added the uncommented out code from the local commit on gocdb-test.

Currently running [here](https://gocdb-www-preprod.gridpp.rl.ac.uk/portal/).

PR #186, opened 2019-10-25
Use new EGI Check-in entitlement to access GOCDB

gregcorbett:
As urn:mace:egi.eu:aai.egi.eu:gocdb is being replaced in favour of an AARC-G027 [compliant one](https://aarc-project.eu/guidelines/aarc-g027/).

Running [here](https://gocdb-www-preprod.gridpp.rl.ac.uk/portal/).

### DIFF
--- a/lib/Authentication/AuthTokens/ShibAuthToken.php
+++ b/lib/Authentication/AuthTokens/ShibAuthToken.php
@@ -133,7 +133,7 @@ class ShibAuthToken implements IAuthentication {
             }
 
             $entitlementValuesArray = explode(';', $_SERVER['entitlement']);
-            if( !in_array('urn:mace:egi.eu:aai.egi.eu:gocdb', $entitlementValuesArray) ){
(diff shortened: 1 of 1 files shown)


lib/Authentication/AuthTokens/ShibAuthToken.php
```diff
@@ -146,7 +146,26 @@ private function getAttributesInitToken(){
             return;
 
         }
-
+        else if($idp == 'https://aai-dev.egi.eu/proxy/saml2/idp/metadata.php'){
```
tofu-rocketry: There's no explanation of why there's all this extra code required when everything else just refers to using a new entitlement.

gregcorbett: This pull request is still a work in progress

tofu-rocketry: Yeah, I've just seen #184. 🤦‍♂

PR #188, opened 2019-11-04
Remove old function from service service

GRyall:
Remove a commented out function from the service service (which is
already overly long and complex).


gregcorbett:
LGTM but there is a conflict :(, can you resolve @GRyall ?

GRyall:
> 
> 
> LGTM but there is a conflict :(, can you resolve @GRyall ?

Done

PR #196, opened 2019-12-19
Refactor Write API request processing

GRyall:
Refactor the getRequestContent function:
* sort out indentation
* simplify some of the logic

### DIFF
--- a/htdocs/PI/write/PIWriteRequest.php
+++ b/htdocs/PI/write/PIWriteRequest.php
@@ -198,43 +198,58 @@ class PIWriteRequest {
     $requestArray = json_decode($requestContents,true,2);
 
     #json_decode returns null for invalid JSON and $requestContents will be empty if there was no request body
-    if(!is_null($requestArray)) {
-      /*The general case expects a single value to be given. In the specific case
-       * of updating extension properties where no specific property is specified
-       * in the url (and so now in $this->entityPropertyKey) we expect multiple
-       * key/value pairs. We use $this->entityPropertyValue and $this->entityPropertyKVArray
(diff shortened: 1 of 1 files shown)


GRyall:
I've just done a little more refactoring to simplify the logic some more. It is easier to read as two commits, so I have left it that way

GRyall:
Final reworking of all the commits into something reviewable

tofu-rocketry:
What was the problem with that reworked logic, and how did you fix it?

GRyall:
> 
> 
> What was the problem with that reworked logic, and how did you fix it?

I missed a valid case for a variable to be null when reworking the function to be less complex. I have now accounted for that case.

htdocs/PI/write/PIWriteRequest.php
```diff
@@ -179,64 +179,78 @@ private function getAndProcessURL($method, $requestUrl) {
...
-            "The JSON message is not correctly formatted. " . $genericError
+            "A value for \"$this->entityProperty\" should be provided. " .
+            "This should be provided in a JSON string {\"value\":\"<value for " .
+            "\"$this->entityProperty\">\"}, with no other pairs present." . $genericError
           );
         }
-    }
+
+        //If there are additional entiries in our array to the value, throw exception
+        if(count($requestArray)==1) {
```
gregcorbett: Seems this should not be `if(count($requestArray)==1)`, greater than 1? not equal to one?

GRyall: yes

GRyall: resolved in latest commit

PR #197, opened 2020-02-03
Update Travis build file

tofu-rocketry:
- Move from Ubuntu Precise to Trusty as base build, keeping Precise for
PHP 5.3 build.
- Simplify to one sqlite build as it fails.
- Add later versions of PHP to test against, adding currently failing
ones to allowed failures.
- Specify OS explicitly.

Example build here: https://travis-ci.org/tofu-rocketry/gocdb/builds/645386379

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
-# N.B. Ubuntu Precise available on Travis until March 2018
-dist: precise
+os: linux
+dist: trusty
 language: php
 before_install:
     - mysql -e 'CREATE DATABASE doctrine;'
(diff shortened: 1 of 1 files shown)


gregcorbett:
Might be informative to also include the [nightly](https://docs.travis-ci.com/user/languages/php/#nightly-builds) build of php?

tofu-rocketry:
> Might be informative to also include the [nightly](https://docs.travis-ci.com/user/languages/php/#nightly-builds) build of php?

My feeling is it wouldn't be as PHP 7.2, .3, nor .4 work anyway, so the nightly's only useful if you're using the latest versions.

PR #198, opened 2020-02-03
Add coverage tests and uploading to Codecov

tofu-rocketry:
Logging into [Codedov](https://codecov.io) with GitHub should show you the details.

I've disabled commenting by Codecov, but it should start showing PR statuses eventually (it might need it to be active on the dev branch first).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,3 +35,6 @@ jobs:
     - env: DB=mysql
       php: '7.4'
     fast_finish: true
+
+after_success:
+    - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


gregcorbett:
I may regret this, but can codecov be made aware of php files in the repo we don't run any tests against?

gregcorbett:
looking at it though, lib and test seem like the best folders to test against. So just whitelist them in phpunit.

tofu-rocketry:
I tried and failed with the whitelisting, partly as I can't use the simple command line option as we're pinned to an old version of PHPUnit, so I've opened an issue to do whitelisting at some point, but will leave it for now. This PR gets the basic coverage working at least.

PR #209, opened 2020-03-09
GT-95 Softcoding of service type production<=>monitored service flag relations.

### LINKED ISSUE #156: 'production => monitored' rule should be a property of the ServiceType in the database and not hardcoded.
I want to be able to throw up databases in a quattorized fashion before I tackle this, but if anyone else wants to tackle it sooner I won't stop them.

ineilson:
Draft pull request while doctrine schema update issues are checked and resolved.

Add monitoring-exception-allowed flag to service type table.
Associated GUI changes and cleanups. 
Add instance count column to service type overview table. 
Add related unit tests.

Closes #156

### DIFF
--- a/htdocs/web_portal/views/admin/add_service_type.php
+++ b/htdocs/web_portal/views/admin/add_service_type.php
@@ -6,7 +6,12 @@
         <input type="text" value="" name="Name" class="input_input_text">
         <span class="input_name">Description</span>
         <input type="text" value="" name="Description" class="input_input_text">
-
+        <span class="input_name" style="">
+            <input type="checkbox" name="AllowMonitoringException" value="checked"/>
+            <label for="AllowMonitoringException">
+                Allow services of this type to be in production without monitoring?
+            </label>
+        </span>
         <br />
         <input type="submit" value="Add Service Type" class="input_button">
     </form>
--- a/htdocs/web_portal/views/admin/add_service_type.php
+++ b/htdocs/web_portal/views/admin/add_service_type.php
@@ -16,3 +16,4 @@
         <input type="submit" value="Add Service Type" class="input_button">
     </form>
 </div>
+<?php
--- a/htdocs/web_portal/views/admin/edit_service_type.php
+++ b/htdocs/web_portal/views/admin/edit_service_type.php
@@ -7,7 +7,15 @@
         <span class="input_name">Description</span>
         <input type="text" value="<?php xecho($params['Description']) ?>" name="Description" class="input_input_text">
 
-        <br />
+        <span class="input_name" style="">
+            <input type="checkbox" name="AllowMonitoringException" value="checked"
+                <?php if ($params['AllowMonitoringException'] == TRUE) echo " checked ";?>
+            />
+            <label for="AllowMonitoringException">
+                Allow <?php xecho($params['Name']) ?> services to be in production without monitoring?
+            </label>
+        </span>
+
         <input class="input_input_hidden" type="hidden" name="ID" value="<?php echo $params['ID'] ?>" />
         <br />
         <input type="submit" value="Submit Changes" class="input_button">
--- a/htdocs/web_portal/views/admin/edit_service_type.php
+++ b/htdocs/web_portal/views/admin/edit_service_type.php
@@ -9,7 +9,11 @@
 
         <span class="input_name" style="">
             <input type="checkbox" name="AllowMonitoringException" value="checked"
-                <?php if ($params['AllowMonitoringException'] == TRUE) echo " checked ";?>
+                <?php
+                if ($params['AllowMonitoringException'] == true) {
+                    echo " checked ";
+                };
+                ?>
             />
             <label for="AllowMonitoringException">
                 Allow <?php xecho($params['Name']) ?> services to be in production without monitoring?
--- a/htdocs/web_portal/views/admin/view_service_type.php
+++ b/htdocs/web_portal/views/admin/view_service_type.php
@@ -1,6 +1,7 @@
 <?php
 $name = xssafe($params['Name']);
 $description = xssafe($params['Description']);
+$allowMonitoringException = xssafe($params['AllowMonitoringExtension']);
 $id = $params['ID'];
 $services = $params['Services'];
 $SEsCount= sizeof($services);
@@ -13,7 +14,9 @@ $portalIsReadOnly = $params['portalIsReadOnly'];
     <!--Headings-->
     <div style="float: left; width: 50em;">
         <h1 style="float: left; margin-left: 0em;">Service Type: <?php echo $name?></h1>
-        <span style="clear: both; float: left; padding-bottom: 0.4em;"><?php echo $description ?></span>
+        <span style="clear: both; float: left; padding-bottom: 0.4em;">
+            <br /><?php require_once __DIR__.'/../fragments/serviceTypeInfo.php'; ?>
+        </span>
     </div>
 
     <!--Edit/Delete buttons-->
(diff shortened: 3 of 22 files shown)


lib/Doctrine/entities/ServiceType.php
```diff
@@ -38,8 +38,16 @@ class ServiceType {
     /** @OneToMany(targetEntity="Service", mappedBy="serviceType") */
     protected $services = null;
 
+    /**
+     * An instance of a Service of this ServiceType may
+     * unmonitored while in production.
+     * @Column(type="boolean", nullable=true)
```
GRyall: Do we definitely want to allow null as a value? Would it not be preferable to have a default?

ineilson: This is the reason for the draft pull, I'm revisiting this very line, trying to understand why I had a problem originally.

GRyall: Sorry - I hadn't spotted this was a draft

gregcorbett: A default of `false` does seem sensible.

ineilson: More ways to skin a cat -
`    public function __construct() {
        $this->services = new ArrayCollection();
        $this->allowMonitoringException = FALSE;
    }`
Which one could argue is more readable than the docblock annotation. But I won't, as the problem I had was the schema update tanking.

ineilson:
I was unable to replicate the schema update problem I ran into originally, so have reverted to suggested definition - which was what I first had afaiks :-(

htdocs/web_portal/views/fragments/serviceTypeInfo.php
```diff
@@ -0,0 +1,23 @@
+<table>
```
gregcorbett, line 1: Could you indent this HTML to improve readability?

gregcorbett (changes requested):
one comment so far. Could you also rebase to resolve the conflicts?

htdocs/web_portal/views/fragments/serviceTypeInfo.php
ineilson: Included in rebased code.

gregcorbett (approved):

gregcorbett:
rebasing on top of latest dev

htdocs/web_portal/views/admin/add_service_type.php
```diff
@@ -6,8 +6,14 @@
         <input type="text" value="" name="Name" class="input_input_text">
         <span class="input_name">Description</span>
         <input type="text" value="" name="Description" class="input_input_text">
-
+        <span class="input_name" style="">
```
Sae126V, line 9: Can we remove style attribute if it is unused ?

htdocs/web_portal/views/admin/add_service_type.php
```diff
@@ -6,8 +6,14 @@
...
+            <input type="checkbox" name="AllowMonitoringException" value="checked"/>
```
Sae126V, line 10: Can you add ID(Unique identifier) for INPUT  element?  So that label can be bound to the input

htdocs/web_portal/views/admin/edit_service_type.php
```diff
@@ -7,9 +7,21 @@
         <span class="input_name">Description</span>
         <input type="text" value="<?php xecho($params['Description']) ?>" name="Description" class="input_input_text">
 
-        <br />
+        <span class="input_name" style="">
```
Sae126V, line 10: Same as above https://github.com/GOCDB/gocdb/pull/209/files#r1147444121. ?

htdocs/web_portal/views/admin/edit_service_type.php
```diff
@@ -7,9 +7,21 @@
...
+            <input type="checkbox" name="AllowMonitoringException" value="checked"
```
Sae126V, line 11: Same as above https://github.com/GOCDB/gocdb/pull/209/files#r1147446474?

htdocs/web_portal/views/admin/view_service_type.php
```diff
@@ -63,32 +78,35 @@
...
                         </td>
 
                         <td class="site_table">
                         <?php
-                        switch($se->getProduction()) {
+                        switch ($se->getProduction()) {
                             case true:
                                 ?>
-                                <img src="<?php echo \GocContextPath::getPath()?>img/tick.png" height="22px" style="vertical-align: middle;" />
+                                <img src="<?php echo \GocContextPath::getPath()?>img/tick.png"
```
Sae126V, line 102: Can we add `alt`  attribute . Just in case, if image is unable to load?

htdocs/web_portal/views/admin/view_service_type.php
```diff
@@ -63,32 +78,35 @@
...
+                                            height="22px" style="vertical-align: middle;" />
                                 <?php
                                 break;
                             case false:
                                 ?>
-                                <img src="<?php echo \GocContextPath::getPath()?>img/cross.png" height="22px" style="vertical-align: middle;" />
+                                <img src="<?php echo \GocContextPath::getPath()?>img/cross.png"
```
Sae126V, line 108: Same as https://github.com/GOCDB/gocdb/pull/209/files#r1147452018 ?

htdocs/web_portal/views/admin/add_service_type.php
gregcorbett: `style=""` is used to a limited extent in other parts of the codebase - seemingly limited to pages around ServiceGroups.

We should determine the effect of removing `style=""` from the codebase and do it in one go if that's determined to be the best thing. Otherwise, we should add comments why the empty style block is needed.

See #445

htdocs/web_portal/views/admin/add_service_type.php
gregcorbett: What is the "label" in this case?

Sae126V: `ID` helps us to identify the element uniquely in the DOM. So, When we are using `label` with `for`. It helps us make a connection between those. For instance, Say another developer declared with the same name as name="AllowMonitoringException. Now, We have two input fields with name="AllowMonitoringException" our `label` field will be confused to make a connection between them. Hope it make sense?

Sae126V: Found this helpful: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

gregcorbett: ah okay, I wasn't seeing the line below this one where `label for` was being used.

What happens if we have the same name but different IDs? Or is that against convention/good html practices - i..e `name`, `label` and `id` should be the same?

Sae126V: IMO: yes, Greg. Because it helps the browser when JS code is loaded as well as the users, usually, the ID and label name should be the same. And the name for the input field helps us to fetch and play with the value(Under the hood: `Name` implies `form control`). 
To answer whether name, label and id should be the same?
ID and label name should be the same always. The name helps to watch or change the form control name(Input field in our case). Ideally, we should NOT use the same input name because it might create an issue when it is performing a lookup for the browser. (We can handle that case with form arrays if the input name is the same). Does it make sense again?

Sae126V: This comment is just meant to indicate that I have updated my comments for better readability.

gregcorbett: thanks for the clarification on this 😄, I think we should address it across the codebase as a whole. @Sae126V can you open a seperate issue to track this?

Sae126V: Cool. Code is looking good to me. So, I am resolving the issues then.

gregcorbett: Tracked in #450

gregcorbett (approved):

PR #215, opened 2020-04-23
Allow access URL based changes to UI menus, colours and general config parameters  

### LINKED ISSUE #216: Allow UI "views" driven by URL-specific configuration.
Allow the UI to be modified based on alternate configuration options keyed on the access URL used.

ineilson:
A small chain of commits allowing modification of behaviour through the local_info.xml file keyed off the URL used to access the site. Closes #216

### DIFF
--- a/config/local_info.xml
+++ b/config/local_info.xml
@@ -85,7 +85,7 @@
             <backgroundDirection>to bottom left</backgroundDirection>
             <backgroundColour1>#F7F9FE</backgroundColour1>
             <backgroundColour2>#DEE9FB</backgroundColour2>
-            <backgroundColour3></backgroundColour3>
+            <backgroundColour3>#D0DFF9</backgroundColour3>
             <headingTextColour>#00AC00</headingTextColour>
           </css>
           <!-- name_mapping
--- a/config/local_info.xml
+++ b/config/local_info.xml
@@ -39,14 +39,14 @@
             ## enable the menu display.
           -->
           <menus>
-            <Projects>true</Projects>
-            <NGI>true</NGI>
+            <Projects>show</Projects>
+            <NGI>show</NGI>
             <AddNGI>show</AddNGI>
             <AddProject>show</AddProject>
-            <ServiceGroups>true</ServiceGroups>
-            <AddSite>true</AddSite>
-            <Scopes>true</Scopes>
-            <AdminScopes>true</AdminScopes>
+            <ServiceGroups>show</ServiceGroups>
+            <AddSite>show</AddSite>
+            <Scopes>show</Scopes>
+            <AdminScopes>show</AdminScopes>
           </menus>
           <!-- css
             ## Provide hex colour values for the background as implemented in
--- a/config/local_info.xml
+++ b/config/local_info.xml
@@ -13,34 +13,7 @@
         -->
         <local_info url="host-172-16-113-107.nubes.stfc.ac.uk">
 
-          <read_only>true</read_only>
-          <show_map_on_start_page>false</show_map_on_start_page>
-
-                  <!-- <filter_downtimes_by_default_scope>true</filter_downtimes_by_default_scope> -->
-
-          <menus>
-            <Projects>hide</Projects>
-            <ServiceGroups>hide</ServiceGroups>
-            <NGI>hide</NGI>
-            <AddNGI>false</AddNGI>
-            <AddProject>false</AddProject>
-            <AddSite>false</AddSite>
-            <Scopes>false</Scopes>
-            <AdminScopes>hide</AdminScopes>
-          </menus>
-
-          <css>
-            <backgroundTop>B5892D</backgroundTop> <!-- // eosc-hub logo brown -->
-            <backgroundBottom>FCF2DD</backgroundBottom>
-            <headingTextColour>#1D3046</headingTextColour> <!-- // eosc-hub logo dark blue -->
-          </css>
-
-          <name_mapping>
-            <Service>
-              <!-- <ServiceEndpoint>Service Component</ServiceEndpoint> -->
-              <endpoint>component</endpoint>
-            </Service>
-          </name_mapping>
+          <default_filter_by_scope>true</default_filter_by_scope>
 
         </local_info>
         <!--
@@ -139,6 +112,11 @@
         <default_scope>
             <name>EGI</name>
         </default_scope>
+
+        <!-- If set to true the initial state of the scope filter drop-downs on the
+             ngi,site,service, downtimes pages is selected for the default_scope
+        -->
+        <default_filter_by_scope>true</default_filter_by_scope>
 
         <!-- Show map on the start page -->
         <show_map_on_start_page>true</show_map_on_start_page>
--- a/config/local_info.xml
+++ b/config/local_info.xml
@@ -1,109 +1,140 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<schema>
-        <local_info>
-        <!-- If read-only is set as true, the portal will prevent changes to the
-             data by all users. This can be used for a read-only failover instance. -->
-        <read_only>false</read_only>
+<config>
+    <!--
+      ## Optional: Begin local override.
+      ##
+      ## One or more local_info elements, with a single url attribute each, can
+      ## optionally be provided to override default configuration values for requests
+      ## arriving from the specified url.
+      ## Multiple local_info elements with at most one url attribute each are supported.
+      ## A single local_info element without url attribute must be specified to provide
+      ## default values for all available configuration elements (i.e. for requests where
+      ## no matching url is found or none are provided).
+      ## Any valid configuration element can be specified here but it MUST exist in
+      ## the default section that follows.
+      ## e.g. Set GOCDB to be read-only and set the default scope to "BIG_PROJECT" -
 
-        <!-- Specify the URLs used to access your GOCDB web portal and PI.
-        - It is important to update these values to match your deployment.
-        - These values substitute placeholder variables and is used in both the
-          PI XML output and in constructing web portal links.
-        - The '/portal' path fragment must match your WEB PORTAL ALIAS mapping
-          in your apache httpd config (see sample 'gocdbssl.conf' file).
-        - The '/gocdbi' path fragment must match your PUBLIC and PRIVATE PI mappings
-          in your apache httpd config (see sample 'gocdbssl.conf' file).
-        - Do not terminate these values with a '/' char.
-        -->
-        <web_portal_url>https://localhost/portal</web_portal_url>
-        <pi_url>https://localhost/gocdbpi</pi_url>
-        <server_base_url>https://localhost</server_base_url>
-        <write_api_user_docs_url>https://wiki.egi.eu/wiki/GOCDB/Write_API/Technical_Documentation</write_api_user_docs_url>
+    <local_info url="REPLACE WITH URL TO MATCH">
 
-        <optional_features>
-            <!-- If false then a service must belong to a physical
-                 site. If true an option will be available to add a brand new
-                 SE to a service group and the new SE won't belong to a
-                 physical site (****MUST be false, true IS NOT SUPPORTED YET****). -->
-            <siteless_services>false</siteless_services>
-        </optional_features>
+      <read_only>true</read_only>
+      <default_scope>
+        <name>BIG_PROJECT</name>
+      </default_scope>
 
-        <!-- default_scope
-        Optional - defines the default scope of the instance.
-        If a default scope IS provided, it is applied when PI queries
-        do not explicitly specify the 'scope' parameter e.g. 'get_site' would return
-        all sites that define the default scope.
-        If default scope is NOT provided, PI queries that do not explicitly specify
-        the 'scope' parameter will return all results regardless of scope, e.g.
-        'get_site' would return all sites in all scopes.
-        -->
-        <default_scope>
-            <name>EGI</name>
-        </default_scope>
+    </local_info>
 
-        <!-- Show map on the start page -->
-        <show_map_on_start_page>true</show_map_on_start_page>
+      ## End local override
+    -->
 
-        <!-- default_scope_match
-        Define the default value for the 'scope_match' PI parameter.
-        Values can be 'any' or 'all'. The default value is applied when multiple
-        scopes are specified in a PI query using the 'scope' parameter and when the
-        'scope_match' PI parameter is not explicitly specified. For example,
-        in this query: 'getSite&scope=EGI,EUDAT,SCOPEX' should sites be returned
-        that define all or any of the specified scopes?
-        -->
-        <default_scope_match>all</default_scope_match>
+    <!--
+      ## Begin defaults. One default element must be present.
+    -->
+    <local_info>
 
-        <!-- Define the minimum number of OPTIONAL scopes required for each IScopedEntity,
-        Optional scopes are NOT in the reserved_scopes list.
-        (Should be positive integer or zero). -->
-        <minimum_scopes>
-            <ngi>1</ngi>
-            <site>1</site>
-            <service>1</service>
-            <service_group>1</service_group>
-        </minimum_scopes>
+      <!-- If read-only is set as true, the portal will prevent changes to the
+          data by all users. This can be used for a read-only failover instance. -->
+      <read_only>false</read_only>
 
-        <!--
-        Specify a list of RESERVED scope tag names:
-        - New Reserved tags can only be directly assigned to resources by the gocdb-admins
-         (resources include NGIs, Sites, Services, SGs but not Projects - don't currently need to tag projects)
-        - When creating a new child resource (e.g. a child Site or child Service),
-          the scopes that are assigned to the parent are automatically inherited and assigned to the child.
-        - Reserved tags assigned to a resource are optional.
-        - Users can reapply Reserved tags to a resource ONLY if the tag can be
-          inherited from the parent ScopedEntity (parents include NGIs/Sites).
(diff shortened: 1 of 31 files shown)


config/local_info.xml
```diff
@@ -1,9 +1,80 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<schema>
+<config> 
+        <!-- 
+          ## Begin local override.
+          ## Specify a url attribute to be matched  against the incoming requestor
+          ## url. Only a single url is supported.
```
gregcorbett, line 6: > Only a single url is supported.

Does that mean only one `<local_info url=X>...</local_info>` block can be defined? 
Or does it mean you can't do things like this `<local_info url=X, Y>...</local_info>`

lib/Gocdb_Services/Config.php
```diff
@@ -87,21 +114,113 @@ public function GetSchemaXML() {
...
      * @return simple xml object
      */
     private function GetLocalInfoXML() {
-        return simplexml_load_file($this->getLocalInfoFileLocation());
+        if ($this->local_info_xml == NULL) {
+            if (!($this->local_info_xml = $this->readLocalInfoXML($this->getLocalInfoFileLocation(), $this->local_info_override))) {
+                throw new \ErrorException("Failed to load xml configuration file: ".$this->getLocalInfoFileLocation());
+            } 
+        } 
+        return $this->local_info_xml;
```
gregcorbett, line 126: So I suppose in some sense, `GetLocalInfoXML` returns what it always returned - a "complete" set of local info. 

It just returns a different set based on the value of `$this->local_info_override`.

Is that right?

config/local_info.xml
```diff
@@ -1,9 +1,80 @@
...
+          <menus>
+            <Projects>true</Projects>
+            <NGI>true</NGI>
+            <AddNGI>show</AddNGI>
+            <AddProject>show</AddProject>
+            <ServiceGroups>true</ServiceGroups>
+            <AddSite>true</AddSite>
+            <Scopes>true</Scopes>
+            <AdminScopes>true</AdminScopes>
+          </menus>
```
gregcorbett, line 50: In this `menus` block, does the use of `true` and `show` illustrate that anything (other than `hide` ,`false`, `0` or `no`) could be used?

I kinda prefer `show`, as this block defines what menus are shown.

config/web_portal/menu.xml
```diff
@@ -59,29 +59,29 @@
 	  <show_on_instance>write_enabled</show_on_instance>
     </spacer>
     
-    <Add_site>
+    <AddSite>
```
gregcorbett, line 62: what was the logic for making these changes to `menu.xml`

config/local_info.xml
```diff
@@ -1,9 +1,80 @@
...
+            ## docs\web_portal\css\web_portal.php).
+            ## [Note that these values are used as part of the dynamic css
+            ## style sheet fetched by the browser after the base html is
+            ## returned.]
+          -->
+          <css>
+            <backgroundTop>F7F9FE</backgroundTop>
+            <backgroundBottom>D0DFF9</backgroundBottom>
+            <headingTextColour>00AC00</headingTextColour>
+          </css>
```
gregcorbett, line 63: The colours in the `css` block dont seem to be the same as the current colours? (comparing https://172.16.112.125/portal/ to https://gocdb-preprod.egi.eu/portal/)

htdocs/web_portal/css/web_portal.php
```diff
@@ -1,3 +1,22 @@
...
+require_once __DIR__ . '/../../../lib/Gocdb_Services/Factory.php';
+header("Content-type: text/css"); 
+// Load variable values from local configuration //
+\Factory::getConfigService()->setLocalInfoOverride($_SERVER['SERVER_NAME']);
+
+$background_top = \Factory::getConfigService()->getBackgroundTop();
+$background_bottom = \Factory::getConfigService()->getBackgroundBottom();
+$header_text_colour = \Factory::getConfigService()->getHeadingTextColour();
+
+/* if (\Factory::getConfigService()->getDefaultScopeName() == 'EOSC-hub') {
```
gregcorbett, line 11: Is there a reason to include this commented out code?

lib/Gocdb_Services/Config.php
```diff
@@ -377,6 +377,22 @@ public function getReservedScopeList() {
         }
         return $reservedScopes;
     }
+    public function getFilterDowntimesByScope () {
```
gregcorbett, line 380: why the special case for downtimes?

lib/Gocdb_Services/Downtime.php
```diff
@@ -738,62 +738,39 @@ public function getDowntimesFilterByParams($filterParams){
         $downtimes = $getDowntime->executeQuery();
         return $downtimes;
     }
-
     /**
-     *
-     * @param \Date $windowStart
-     * @param \Date $windowEnd
      */
-    public function getActiveAndImminentDowntimes($windowStart, $windowEnd) {
```
gregcorbett, line 747: so `getActiveAndImminentDowntimes` was never used it seems...

gregcorbett:
One thing I’ve noticed is that when editing a site with this code you only get the option of applying the default scope or the local scope. It also seems like the SLA scope tag disappears for some reason?

For example:

https://host-172-16-112-125.nubes.stfc.ac.uk/portal/index.php?Page_Type=Edit_Site&id=335 (url based override in place)
215-url-match

https://172.16.112.125/portal/index.php?Page_Type=Edit_Site&id=335 (same host as above, no url based override in place)
215-default

https://gocdb-preprod.egi.eu/portal/index.php?Page_Type=Edit_Site&id=335 (current behaviour)
gocdb-preprod

In the case when no url based override in place, we'd want to be able to apply any scope tag.

config/local_info.xml
ineilson: No. Multiple local_info element blocks are allowed. And yes, you can't specifiy multiple values per url attribute.

lib/Gocdb_Services/Config.php
ineilson: Correct. In fact, in the xml file there must be one (and only one) <config> element without a url= attribute and this must be 'complete', meaning all sub-elements in the schema must be given a value there.  Defining a <config url= ... > effectively overwrites the values from this 'base' element if the url is matched.

config/local_info.xml
ineilson: Yes. Basically anything else shows the menu. Not sure if that's best really. I prefer show and hide and my guess is if you distributed the 'first' version of the file with an example using only these values, they'd become the de facto standard.

config/web_portal/menu.xml
ineilson: That's a good question. It was it impose some consistency so that the values here could be programatically converted to the local_info.xsd schema definition. So, a commit crossover, and needs a bit more investigation to figure out if I've missed something in refactoring to separate commits.

config/local_info.xml
ineilson: Laziness I'm afraid. Comparing the 'new' web_portal.php with the original plain .css, I used a 2-way linear-gradient only. I guess I thought the difference slight. (screenshot as getting this to render in markdown was beyond my patience.) Could be changed obviously with the addition of more config parameters.

image

htdocs/web_portal/css/web_portal.php
ineilson: None known now. I've removed it.

lib/Gocdb_Services/Config.php
ineilson: Not sure exactly what you ask, and it's been a long time since, but AFAIR discussions, the display/filtering of downtimes was inconsistent compared to other entities in being filtered by default. Introducing the new config flag filter_downtimes_by_default_scope especially for downtimes allows, if set to "true", to mimic existing behaviour. If set to false, downtimes, along with everything else, are only filtered if the default_filter_by_scope flag is true.

lib/Gocdb_Services/Downtime.php
ineilson: It seems ...

config/local_info.xml
gregcorbett: Indeed, could you use just `show` here then?

gregcorbett:
> It also seems like the SLA scope tag disappears for some reason?

It turns out this is a something to do with the quattorization, one for me to fix.

config/local_info.xml
gregcorbett: ah I see, could you add the additional config parameters?

config/local_info.xml
ineilson: OK. 
Updated.

ineilson:
> One thing I’ve noticed is that when editing a site with this code you only get the option of applying the default scope or the local scope. It also seems like the SLA scope tag disappears for some reason?
> 
> For example:
> 
> https://host-172-16-112-125.nubes.stfc.ac.uk/portal/index.php?Page_Type=Edit_Site&id=335 (url based override in place)
> 215-url-match
> 
> https://172.16.112.125/portal/index.php?Page_Type=Edit_Site&id=335 (same host as above, no url based override in place)
> 215-default
> 
> https://gocdb-preprod.egi.eu/portal/index.php?Page_Type=Edit_Site&id=335 (current behaviour)
> gocdb-preprod
> 
> In the case when no url based override in place, we'd want to be able to apply any scope tag.

Have reverted to original behaviour. I don't know why this was introduced.

gregcorbett:
> It turns out this is a something to do with the quattorization, one for me to fix.

Fixed.

config/local_info.xml
```diff
@@ -37,10 +108,16 @@
...
         -->
+
         <default_scope>
             <name>EGI</name>
         </default_scope>
 
+        <!-- If set to true the initial state of the scope filter drop-downs on the
+             ngi,site,service, downtimes pages is selected for the default_scope
+        -->
+        <default_filter_by_scope>true</default_filter_by_scope>
```
gregcorbett, line 119: Shouldn't this be `false` to mimic existing behaviour?

ineilson: Yes, I believe so.

config/local_info.xml
ineilson: Updated to allow more generalised colour spec., including diagonal gradient.

config/web_portal/menu.xml
ineilson: So ...
Commit https://github.com/ineilson/gocdb/commit/e9487dff699ef2f9778a32d28ddaa1f05dcbe498 contains .xslt code to  generate the menu element specifications which are included in the .xsd file. The XML schema enforcement/checking is not yet in the dev branch so menu elements (and colour spec.) elements don't exist yet to be validated. After the XML schema commit is merged then the present commit will need updating with the required XML schema elements, as per the commit above.

lib/Gocdb_Services/Config.php
gregcorbett: following some investigation, @ineilson and I had trouble recreating the existing behaviour for both the `Downtimes_Calendar` page and the `Sites` page .

it seems as if the two filtering options (`default_filter_by_scope` and `filter_downtimes_by_default_scope`) may be linked in some way (if one is false the other is always false perhaps) and I dont think they should be.

config/web_portal/menu.xml
gregcorbett: I'm a little confused... is ineilson@e9487df on an older branch where the URL specific views and schema checking were more intertwined?

ineilson: Yes, that sums it up. Whichever branch (this filter-by-scope or local_info-schema-checking) is merged first, the other will need some changes.

config/local_info.xml
gregcorbett:
```suggestion
        <default_filter_by_scope>false</default_filter_by_scope>
```

lib/Gocdb_Services/Config.php
ineilson: Summary of offline discussions: decided to treat the current Downtimes_Calendar page behaviour (i.e. silently filtering by `default scope`) as a bug and make all pages respond consistently to the setting of `default_filter_by_scope`, eliminating the need for `filter_downtimes_by_default_scope`.

config/local_info.xml
ineilson: Updated in latest commit - https://github.com/GOCDB/gocdb/blob/1b51b69234b9827a75535fc6fdd67d5a9c1006a9/config/local_info.xml#L150

lib/Gocdb_Services/Config.php
ineilson: `filter_downtimes_by_default_scope` removed in 1b51b69234b9827a75535fc6fdd67d5a9c1006a9

htdocs/web_portal/views/downtime/downtimes_overview.php
```diff
@@ -25,11 +26,17 @@
         </div>
         <div style="float: left;">
             <h1 style="float: left; margin-left: 0em; padding-bottom: 0.0em;">
-                   Overview of Current and Planned Downtimes
+                   Downtimes
```
gregcorbett, line 29: Can you revert this line? I think it's clearer as before.

```suggestion
                   Overview of Current and Planned Downtimes
```

htdocs/web_portal/views/downtime/downtimes_overview.php
```diff
@@ -25,11 +26,17 @@
...
-                All currently active and planned downtimes over the coming weeks<br />
-             </span>
+                Currently active and scheduled downtimes<!-- This comment removes the pad space
+                --><?php
+                    if ($filterScope != NULL) {
+                        echo ', filtered by scope "' . $filterScope . '"';
+                    }
+                ?><!-- This comment removes the pad space
+                -->.<br />
+            </span>
```
gregcorbett, line 39: I'm torn on this change, but I think i prefer it as before. Could you revert this chunk?

gregcorbett:
Just to track it, the monitoring page is still not handling the multiple local_info blocks correctly, it's just picking the first one.

htdocs/web_portal/views/downtime/downtimes_overview.php
ineilson: Reverted.

htdocs/web_portal/views/downtime/downtimes_overview.php
ineilson: Reverted. I must kick the habit of trying to make things better ;-)

ineilson:
> * If you got to https://host-172-16-113-107.nubes.stfc.ac.uk/portal/index.php?Page_Type=Sites, uncheck EOSChub in the "Site Scopes" box and hit "Filter Sites", you end up still filtering by "EOSChub".

For the case where default_filter_by_scope is true -

Correct, but the uncheck leaves no scope selected, so what do we do? - apply a filter selecting no scopes (= empty result) or drop back to the default behaviour (= select default scope)? The former is logical behaviour and what the user asked for, but also somewhat pointless. I think I'd prefer an error dialogue "Select at least one scope."

Have I got this wrong?

gregcorbett:
I would say we should default to the current behavior as much as possible, so if you manual deselect all the scopes and hit "Filter Sites", you should get all Sites.

ineilson:
> * When adding an entity, the default scope of the portal should be pre-selected (and the `local` scope)

This breaks the inheritance scheme, carefully implemented, where the default scope(s) for an entity (service or site) are taken from the selected parent (site or ngi respectively). It could, presumably be done for creating/editing an ngi or service group, which have no parent.

GRyall:
> 
> 
> > * When adding an entity, the default scope of the portal should be pre-selected (and the `local` scope)
> 
> This breaks the inheritance scheme, carefully implemented, where the default scope(s) for an entity (service or site) are taken from the selected parent (site or ngi respectively). It could, presumably be done for creating/editing an ngi or service group, which have no parent.

I think this needs further discussion and a decision to be made one way or the other. presumably if people are viewing the site through a specific scope view and they can see the entities parent then it will have the scope of that view otherwise they wouldn't be able to see it?

gregcorbett:
> 
> 
> > > * When adding an entity, the default scope of the portal should be pre-selected (and the `local` scope)
> > 
> > 
> > This breaks the inheritance scheme, carefully implemented, where the default scope(s) for an entity (service or site) are taken from the selected parent (site or ngi respectively). It could, presumably be done for creating/editing an ngi or service group, which have no parent.
> 
> I think this needs further discussion and a decision to be made one way or the other. presumably if people are viewing the site through a specific scope view and they can see the entities parent then it will have the scope of that view otherwise they wouldn't be able to see it?

Ah okay. If the pre-ticked scope are getting inherited from the scope tags applied to the parent entity, that seems sufficient. But we can talk more later today.

gregcorbett:
Maybe I'm doing something wrong, but the monitoring page still seems to be picking the "first" `{PI|Portal|Server Base} URL` in `local_config.xml`

For example: both these URL views on the same host [1, 2] show the Portal URL as: https://gocdb-preprod-eosc.stfc.ac.uk/portal. 
For [1], that's to be expected, but for [2] I'd expect it to pick the "default".

You should be able to login to host-172-16-112-125.nubes.stfc.ac.uk if you want to eyeball the `local_config.xml`, maybe I've set it up wrong.

[1] https://gocdb-preprod-eosc.stfc.ac.uk/portal/GOCDB_monitor/
[2] https://host-172-16-112-125.nubes.stfc.ac.uk/portal/GOCDB_monitor/

ineilson:
Well, as I started digging, in case you end up back here, I suspected it might be related to the issue nicely described in the most popular answer here - https://stackoverflow.com/questions/2297403/what-is-the-difference-between-http-host-and-server-name-in-php. Can you try using UseCanonicalName in the Apache config?

gregcorbett (approved):

PR #220, opened 2020-04-27
Refactor the write API authentication

GRyall:
Move obtaining authentication details out of the write API class in order to
enable future functional testing of that class

Also a minor addition of some documentation to a function


htdocs/PI/write/PIWriteRequest.php
```diff
@@ -556,17 +556,27 @@ private function checkIfGOCDBIsReadOnlyAndRequestisNotGET(){
 
 
     /**
-     * Gets authentication information and sets the relevant class property
+     * Sets the relevant class property
      */
-    private function getAndSetAuthInfo() {
+    private function getAndSetAuthInfo($authArray) {
```
gregcorbett, line 561: If this no longer gets authentication information, the function should be renamed.

GRyall: Agreed, fixed in latest commit

PR #221, opened 2020-04-29
Allow <> through service extension property validation.

### LINKED ISSUE #214: Input character validation to allow "<>" in endpoint extension properties
Allow "<>" through input validation of extension properties to enable the encoding of multiple host DNs - to be used by a single service offered on multiple hosts (each with unique Subject DN and shared service SubjectAltName) behind load-balanced URL. e.g. -
image

Ref: RFC1779

ineilson:
Including basic ServiceService extension property input validation tests. Fixes #214

### DIFF
--- a/tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
+++ b/tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
@@ -1,0 +1,195 @@
+<?php
+/*
+ * Copyright (C) 2015 STFC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+require_once __DIR__ . '/../../../doctrine/TestUtil.php';
+require_once __DIR__ . '/../../../../lib/Gocdb_Services/ServiceService.php';
+require_once __DIR__ . '/../../../../lib/Gocdb_Services/Factory.php';
+use Doctrine\ORM\EntityManager;
+require_once __DIR__ . '/../../../doctrine/bootstrap.php';
+
+/**
+ * DBUnit test class 
+ *
+ */
+class ServiceServiceTest extends PHPUnit_Extensions_Database_TestCase{
+  private $em;
+
+  /**
+  * Overridden.
+  */
+  public static function setUpBeforeClass() {
+    parent::setUpBeforeClass();
+    echo "\n\n-------------------------------------------------\n";
+    echo "Executing ServiceServiceTest. . .\n";
+  }
+
+  /**
+  * Overridden. Returns the test database connection.
+  * @return PHPUnit_Extensions_Database_DB_IDatabaseConnection
+  */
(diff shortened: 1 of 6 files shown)


gregcorbett:
Have the argus folk confirmed they can parse the escaped "less than" and "greater than" symbols?

ineilson:
Well, not exactly, claim: "could deal with any format" -

> I haven't had time to try to really think about the use of &lt; and &gt; to separate the values.
> The parser I have could indeed deal with any format, but I am not too sure:
> - If it would impact other consumers of gocdb data (I don't know if anyone else reads this)
> - If there are no "standard" for such thing
> The good thing is that I would not expect '&gt;' to be part of a valid DN, so we wouldn't have to deal with escaping the DN.

For the above bullet points we have already discussed and decided to go ahead. For further testing I thought we were going to put this on a/the test server and advertise the change for people to know and check. There is also a plan to raise this as part of the next security team meeting.

tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Check the basics -
+   * Duplicates some simple testing from ExtensionsTest
+   */
+  public function testValidateProperty(){
+    print __METHOD__ . "\n";
+
+    list ($service, $user, $serviceService) = $this->createValidationEntities();
+   
+    $values[0] = array("ValidName1","ValidValue");
```
gregcorbett, line 149: are these arrays of the form ("Property Name", Property Value")?

tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
```diff
@@ -0,0 +1,195 @@
...
+   * Check that validation of property name is operating as expected
+   * Added to test code changes to pass through "<>" chars
+   * @depends testValidateProperty
+   */
+  public function testValidatePropertyNameFails(){
+    print __METHOD__ . "\n";
+
+    list ($service, $user, $serviceService) = $this->createValidationEntities();
+   
+    $values[0] = array("<Invalid>","<valid>");
```
gregcorbett, line 172: Is this array of the form ("Property Name", Property Value")? Its worth a comment as to why this case is invalid.

tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Check that validation of property value is operating as expected
+   * @depends testValidateProperty
+   */
+  public function testValidatePropertyValueFails(){
+    print __METHOD__ . "\n";
+
+    list ($service, $user, $serviceService) = $this->createValidationEntities();
+   
+    $values[0] = array("Valid","'Not Valid'");
```
gregcorbett, line 188: Is this array of the form ("Property Name", Property Value")? Its worth a comment as to why this case is invalid.

tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
ineilson: Yes.

tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
ineilson: Added comment

tests/unit/lib/Gocdb_Services/ServiceServiceTest.php
ineilson: Added comment

gregcorbett (dismissed):

gregcorbett:
@ineilson, could this be rebased and the conflicts fixed?

gregcorbett (approved):

PR #226, opened 2020-06-10
Expand documentation for installing on MariaDB

ineilson:


gregcorbett:
Looks great @ineilson :)

PR #231, opened 2020-07-02
Align body of STFC logo and clean css for sponsor logos.

ineilson:
Aligns the top of the UKRI logo with the top of the other sponsor logos on the landing and in the left menu, allowing the STFC council symbol to project above the upper line, to better balance the visual weight of the logos. Is stable with IE11 so long as compatibility mode is not enabled. Also increases the size of the logos on the landing page slightly to avoid tiny-font syndrome. Some cleaning of the CSS classes and associated code. Added "part of UK Research and Innovation" to landing page acknowledgement.

Before -
image
image

After -
image
image

### DIFF
--- a/htdocs/landing/home.css
+++ b/htdocs/landing/home.css
@@ -88,8 +88,8 @@ a {
 div.Copyright_Text {
     font-size: 0.8em;
     text-align: center;
-    margin-top: 1em;
-    margin-bottom: 1em;
+    margin: 1em auto;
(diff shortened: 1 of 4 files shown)


ineilson:
Added UKRI link, add margins to balance the text and restore colours for links. Once highlighted as links, it seems strange, but minor issue, that both "EGI" and "EGI.eu" point to the same page.
image

PR #242, opened 2020-08-06
Add base directory .htaccess to .gitignore

ineilson:
When testing on a active web-serving instance, this prevents .htaccess showing up as untracked in git.


GRyall:
Do we use .htaccess much?

gregcorbett:
Not consciously

PR #246, opened 2020-08-17
 Allow multiemail input for ServiceGroup EMAIL BUT INTO DEV

GRyall:
contents of #146 into dev rather than master


gregcorbett:
discussing offline the best way to proceed.

PR #249, opened 2020-08-24
Unregistered user protected messages fix

### LINKED ISSUE #248: Info message typo in Service view
Unregistered user sees authentication required messages, some are misspelled.

image

tofu-rocketry:
Resolves #248.

This PR corrects the spelling and also improves the messages, wherever they occur in the codebase, to clarify that registration is required and not just "auth" as you need a grid cert to get that far anyway.

Also makes the PHP logic around one of the instances consistent with the others in the same file.

### DIFF
--- a/htdocs/web_portal/views/ngi/view_ngi.php
+++ b/htdocs/web_portal/views/ngi/view_ngi.php
@@ -54,7 +54,7 @@
                             <a href="mailto:<?php xecho($params['ngi']->getEmail()) ?>">
                                 <?php xecho($params['ngi']->getEmail()) ?>
                             </a>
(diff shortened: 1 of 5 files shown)


jrha (approved):
This change looks good to me.

It feels like a further improvement would be to start defining "standard" error messages like these as constants in a separate file, which should discourage the adding of more non-standard message and possibly improve future portability/customisation/translation.

tofu-rocketry:
> It feels like a further improvement would be to start defining "standard" error messages like these as constants in a separate file, which should discourage the adding of more non-standard message and possibly improve future portability/customisation/translation.

Issue opened at #259 .

gregcorbett (approved):
LGTM

PR #250, opened 2020-08-24
Update to titles for users

### LINKED ISSUE #176: Title on user registration page
The list of titles on the user registration page currently has no gender neutral option (e.g. Mx) unless you happen to be a Dr. If we really do need to capture this, we should add an option. I would also suggest we should allow this field to be nullable. 

It also defaults to Mr, which does not send the right message either - I suggest defaulting to null.

GRyall:
Fixes #176. On user edit and registration pages:

* Add title 'Mx'
* Add option for empty title
* Update Gocdb schema to allow empty string titles (simpler to
maintain than having a null option on the drop down)
* Reorder titles to be in alphabetical order

I have tested the edit page. I have not been able to test the registration page (though the code is very similar). It should be tested before merging.

### DIFF
--- a/config/gocdb_schema.xml
+++ b/config/gocdb_schema.xml
@@ -243,7 +243,7 @@
 	<field>
 	    <fname>TITLE</fname>
 	    <length>255</length>
-	    <regex>/^[a-zA-Z\s\-']+$/</regex>
+	    <regex>/^[a-zA-Z\s\-']*$/</regex>
 	</field>
 	<field>
 	    <fname>CERTIFICATE_DN</fname>
--- a/htdocs/web_portal/views/user/edit_user.php
+++ b/htdocs/web_portal/views/user/edit_user.php
@@ -6,7 +6,7 @@
             Title
         </span>
         <?php
-            $titles = array('Mr', 'Mrs', 'Miss', 'Ms', 'Prof', 'Dr');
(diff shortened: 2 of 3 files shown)


tofu-rocketry (approved):
🏳️‍🌈

config/gocdb_schema.xml
```diff
@@ -243,7 +243,7 @@
 	<field>
 	    <fname>TITLE</fname>
 	    <length>255</length>
-	    <regex>/^[a-zA-Z\s\-']+$/</regex>
+	    <regex>/^[a-zA-Z\s\-']*$/</regex>
```
tofu-rocketry, line 246: Are there any titles we have that match on `\s` or `\-`?

GRyall: No

tofu-rocketry: Fair enough. Was partly wondering about the contents of the database, and if there were any potential future ones that might.

GRyall: Then we can change it when we add the future ones. This is more permissive than what we had before.

tofu-rocketry: Sure, which is why I approved this PR and just added a comment.

gregcorbett (approved):
Registering as a new user works as expected, as does editing a user.

PR #251, opened 2020-08-24
Add styling to read only message

### LINKED ISSUE #65: Read only mode warning appears in wrong place
From the 5.7 release the read only mode warning appears next to, instead of beneath, the version text

tofu-rocketry:
Resolves #65 (which isn't quite 4 years old yet).

This corrects the placement of the message so it appears below the GOCDB + version line and also changes it to all caps to make it more obvious.

Only tested by making edits through Firefox Web Dev tools Inspector, though it only touches HTML so should be safe, but actual testing would be a good idea.


jrha (approved):
LGTM

gregcorbett (approved):
LGTM, and seems to work as expected on both Firefox and Chrome.

Issue #203, opened 2020-02-11
Authentication attributes box on user page shows the AuthenticationRealm of the logged in user, not the user being viewed.


gregcorbett:
This was not the behaviour I was expecting. 

I thought it displayed how the user being viewed logs into GOCDB, but doing some digging I don't believe this information is even stored per user...

GRyall:
For non-admins, the only user page accessible is the one for that user (accessed with the 'view details' link). The page is then reused for the admin view on all users. I suggest we simply hide the box when you are not viewing yourself. 

Given the above, I am labelling this as an enhancement (no bug is visible to end users).

gregcorbett:
> For non-admins, the only user page accessible is the one for that user (accessed with the 'view details' link). 

Non-admin users can view any other user page, so it is a bug visible to end users.

> I suggest we simply hide the box when you are not viewing yourself.

That seems like a sensible fix.

PR #253, opened 2020-08-26
Make style of auth attr box consistent with site style

### LINKED ISSUE #84: Authentication attributes box on user page does not match theme of page
The style of this box does match the rest of the page. From a user perspective, it is unclear what this box means.

image

tofu-rocketry:
Resolves #84.

This fixes the style of the authentication attributes box. It does abuse the table styles somewhat, but an improvement would be to use the PHP to generate individual lines for the table. (Though is there ever more than one line of attributes?)

Also tidies up the indentation somewhat.


GRyall:
This only partially fixes #84. It sorts out the style, but it doesn't make it any clearer to the user what this box is about. A clearer title may fix this to some extent

tofu-rocketry:
Any suggestions? I don't understand it particularly myself.

GRyall:
> 
> 
> Any suggestions? I don't understand it particularly myself.

- "Authentication Attributes" -> "User Authentication Details" or similar
- Remove the square brackets and put the two bits into separate columns?

tofu-rocketry:
@gregcorbett What's your preferred terminology?

gregcorbett:
> Though is there ever more than one line of attributes?

In theory, yes. In practice, no, or at least not yet. They are just arbitrary further "user details" that get set when GOCDB determines how you have logged into your current session (see [X509AuthenticationToken.php](https://github.com/GOCDB/gocdb/blob/dev/lib/Authentication/AuthTokens/X509AuthenticationToken.php#L29))

>    * "Authentication Attributes" -> "User Authentication Details" or similar
>    * Remove the square brackets and put the two bits into separate columns?

That looks like a good solution to address the rest of #84.

A better solution could be to address #203 at the same time and show these extra details in the "User Details" table that already exists.

tofu-rocketry:
> A better solution could be to address #203 at the same time and show these extra details in the "User Details" table that already exists.

If we're going to hide this for non-current users, then I'd say we should keep it as a separate box rather than changing how many elements appear in the User Details table.

gregcorbett (changes requested):
It seems the auth details table doesn't show when self viewing, and stops the subsequent role tables from displaying (both when self viewing and viewing another user)?

image

htdocs/web_portal/controllers/user/view_user.php
```diff
@@ -79,9 +79,11 @@ function view_user() {
                     $r->setDecoratorObject('GOCDB ADMIN');
                 }
             }
+            $params['viewingSelf'] = false;
     } else {
         // current user is viewing their own roles, so they can revoke their own roles
         $r->setDecoratorObject('[Self revoke own role]');
+        $params['viewingSelf'] = true;
```
gregcorbett, line 86: I would separate this out from the block of code that determines if the user can revoke roles of the user they are looking at.

tofu-rocketry: Yes, that's fixed some of the weirdness!

tofu-rocketry:
Finally fixed it! I am not a fan of PHP.

The lack on indentation in one place meant I didn't spot that one bit would only run if a user had roles, then I tried to use a variable before it was set, then I forgot to use the comparison operator. Wasn't clear what was going wrong any of those times!

Anyway, here's how it looks for the current user. Box is correctly hidden when view another user.

image

gregcorbett:
Sorry, I think #294 has superceeded this PR by resolving both https://github.com/GOCDB/gocdb/issues/84, and resolves https://github.com/GOCDB/gocdb/issues/203.

PR #260, opened 2020-12-08
IAM Authentication

sarahbyrnie:
Integration with the IRIS IAM service. All changes can be viewed on https://host-172-16-114-81.nubes.stfc.ac.uk/


lib/Gocdb_Services/Site.php
```diff
@@ -1408,6 +1408,11 @@ public function addAPIAuthEntity(\Site $site, \User $user, $newValues) {
         if ($type == 'X509' && !preg_match("/^(\/[A-Za-z]+=[a-zA-Z0-9\/\-\_\s\.,'@:\/]+)*$/", $identifier)) {
             throw new \Exception("Invalid x509 DN");
         }
+        
+        //If the entity is of type OIDC subject, do a more thorough check again
+        if ($type == 'OIDC Subject' && !preg_match("/^([a-z0-9]{8}\-[a-z0-9]{4}\-[a-z0-9]{4}\-[a-z0-9]{4}\-[a-z0-9]{12})$/", $identifier)) {
```
norealroots, line 1413: I was asked about whether this format would change - for completeness, this is the standard format for a Universally Unique identifier (UUID): [https://en.wikipedia.org/wiki/Universally_unique_identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)

htdocs/landing/index.php
```diff
@@ -21,7 +21,27 @@ <h1 class="Landing_Welcome">Welcome to GOCDB</h1>
             <p>Use of GOCDB is governed by the <a class="docLink hover" href="aup.html">EGI Acceptable Use Policy</a> which places restrictions on your use of the service.</p>
             <p>The <a class="docLink hover" href="privacy.html">GOCDB Privacy Notice</a> describes what personal data is collected and why, and your rights regarding this data.</p>
             <p> Please read these documents before accessing GOCDB.</p>
-            <a href="/portal/" class="button">Access GOCDB</a>
+          </div>
+          <div style="width: 80%; margin-left: auto; margin-right: auto;">
+            <a href="/portal/" style="width:73%; font-size:1.7em" class="button">Access GOCDB via X.509 Certificate</a>
```
ineilson, line 26:
```suggestion
            <a href="/portal/" style="width:73%; font-size:1.7em" class="button">Access GOCDB using your IGTF X.509 Certificate</a>
```

htdocs/landing/index.php
```diff
@@ -21,7 +21,27 @@ <h1 class="Landing_Welcome">Welcome to GOCDB</h1>
...
+                $hostname = $_SERVER['HTTP_HOST'];
+                $egi_target = urlencode("https://" . $hostname . "/portal/");
+                $egi_redirect = "https://" . $hostname . "/Shibboleth.sso/Login?target=" . $egi_target;
+                if($_SERVER['REQUEST_URI'] === "/"){
+                    $iam_target = "target_link_uri=" . urlencode("https://" . $hostname . "/portal/");
+                }
+                else{
+                    $iam_target=ltrim($_SERVER['REQUEST_URI'], '/?');
+                }
+                $iam_redirect = "https://" . $hostname . "/portal/redirect_uri?iss=https%3A%2F%2Firis-iam.stfc.ac.uk%2F&" . $iam_target;
```
ineilson, line 40: Why is it not possible to use urlencode() here?

htdocs/web_portal/components/Get_User_Principle.php
```diff
@@ -162,21 +162,29 @@ function Get_User_Principle(){
 }
 
 /**
- * Get the DN from an x509 cert or null if a user certificate can't be loaded.
- * Called from the PI to authenticate requests using certificates only.
+ * Get the DN from an x509 cert, principle from oidc auth, or null if a user certificate can't be loaded.
```
ineilson, line 165:
```suggestion
 * Get the DN from an x509 cert, Principal from oidc token, or null if a user certificate can't be loaded.
```

htdocs/web_portal/controllers/user/register.php
```diff
@@ -56,10 +56,20 @@ function draw() {
     die();
     }
 
+    //Extract users email from oidc claims
+    $authDetails = $_SERVER['OIDC_CLAIM_external_authn'];
```
ineilson, line 60: Not sure how it would work but it feels like the global OIDC_CLAIM_* values are crying out for an accessor class.

lib/Authentication/AuthTokens/IAMAuthToken.php
```diff
@@ -0,0 +1,132 @@
...
+            $this->principal = $_SERVER["REMOTE_USER"];
+            $this->userDetails = array('AuthenticationRealm' => array('IRIS IAM - OIDC'));
+            //Don't allow access if user only has a local account on IRIS
+            if(strpos($_SERVER['OIDC_CLAIM_groups'], "localAccounts")===false){
+            }else{
+                die('You must login via your organisation on IAM to gain access to this site.');
+            }
+            //Don't allow access unless user is a member of the IRIS gocdb group
+            if(strpos($_SERVER['OIDC_CLAIM_groups'], "gocdb")===false){
+                die('You do not belog to the correct group(s) to gain access to this site.');
```
ineilson, line 89:
```suggestion
                die('You do not belong to the correct group(s) to gain access to this site.');
```
Could this be rephrased to help the user? Something: "Contact blah blah to join blah blah groups to access this site."

lib/Authentication/AuthTokens/IAMAuthToken.php
```diff
@@ -0,0 +1,132 @@
...
+    
+    
+    private function getAttributesInitToken(){
+        if(isset($_SERVER['OIDC_access_token'])){
+            $this->principal = $_SERVER["REMOTE_USER"];
+            $this->userDetails = array('AuthenticationRealm' => array('IRIS IAM - OIDC'));
+            //Don't allow access if user only has a local account on IRIS
+            if(strpos($_SERVER['OIDC_CLAIM_groups'], "localAccounts")===false){
+            }else{
+                die('You must login via your organisation on IAM to gain access to this site.');
```
ineilson, line 85:
```suggestion
                die('You must login via your organisation on IRIS IAM to gain access to this site.');
```

lib/Gocdb_Services/Site.php
sarahbyrnie: Looking at the standard format it should only contain characters a-f not a-z, so I think the following is better to avoid a 'false positive'

```suggestion
        if ($type == 'OIDC Subject' && !preg_match("/^([a-f0-9]{8}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{12})$/", $identifier)) {
```

norealroots: Good point - missed that! All characters in the UUID are Hex, so this makes sense.

lib/Authentication/AuthTokens/IAMAuthToken.php
norealroots: It's controlled via IAM, but would probably want to be more generic and so point you to a GOCDB admin rather than an IAM/OIDC admin?

sarahbyrnie: Or maybe just point the user back to IAM to request group membership?

```suggestion
                die('You do not belong to the correct group to gain access to this site. Please visit iris-iam.stfc.ac.uk to request to join the gocdb group.');
```

norealroots: I just thought that because you have `group(s)` it may not be as simple as requesting to join the GOCDB group? Depends if it's one group for everything or you are using subgroups to protect specific access further.

If it is one group, then not sure it's fully necessary, but to me it feels good to say it will need reviewing - ie not an instant process. Also aware it's now a very long message though.

```suggestion
                die('You do not belong to the correct group(s) to gain access to this site. Please visit iris-iam.stfc.ac.uk and submit a membership request to the GOCDB group. This shall be reviewed by a GOCDB admin.');
```

sarahbyrnie: It said group(s) because the gocdb group and the localAccounts  group were originally handled together, but it is actually now just the one group called gocdb being referred to here

htdocs/web_portal/components/Get_User_Principle.php
sarahbyrnie: I missed that it's not just a user cert that can't be loaded

```suggestion
 * Get the DN from an x509 cert, Principle from oidc token, or null if neither can be loaded.
```

htdocs/landing/index.php
sarahbyrnie: Changes the size of the button as well to better format with new text

```suggestion
            <a href="/portal/" style="width:68%; font-size:1.7em" class="button">Access GOCDB using your IGTF X.509 Certificate</a>
```

lib/Authentication/AuthTokens/IAMAuthToken.php
sarahbyrnie: I think the long message is okay, removing the plural.

```suggestion
                die('You do not belong to the correct group to gain access to this site. Please visit iris-iam.stfc.ac.uk and submit a membership request to the GOCDB group. This shall be reviewed by a GOCDB admin.');
```

ineilson: Sorry to nit-pick but, is that submit a request **to** the group or **for** (membership of) the group? Suggest maybe : 
`... submit a request for membership of the GOCDB group for review by an administrator.`  
Is that what's meant?

sarahbyrnie: It's not nit-picking, but definitely testing my grammar skills. I think this is less confusing.

```suggestion
                die('You do not belong to the correct group to gain access to this site. Please visit iris-iam.stfc.ac.uk and submit a request to join the GOCDB group. This shall be reviewed by a GOCDB admin.');
```

htdocs/landing/index.php
sarahbyrnie: Do you mean on the iam_target or iam_redirect?

ineilson: iam_redirect - urlencode("https://iris-iam.stfc.ac.uk/") or similar ?

sarahbyrnie: That bit is I think already urlencoded and just pasted in

ineilson: Granted. It's really a style/readability comment. You're welcome to disagree ;-)

sarahbyrnie: I was a bit worried that the php block was getting too long, but something like this is definitely more readable

```suggestion
                $iris_url = urlencode("https://iris-iam.stfc.ac.uk/");
                $iam_redirect = "https://" . $hostname . "/portal/redirect_uri?iss=" . $iris_url . "&" . $iam_target;
```

htdocs/web_portal/controllers/user/register.php
```diff
@@ -56,10 +56,20 @@ function draw() {
...
+    $startPos = 3+strpos($authDetails, ":", (strpos($authDetails, "MAIL")));
+    $endPos = strpos($authDetails, "\"", 3+$startPos);
+    $length = $endPos-$startPos;
+    $userEmail = substr($authDetails, $startPos, $length);
+
     /* @var $authToken \org\gocdb\security\authentication\IAuthentication */
     $authToken = Get_User_AuthToken();
     $params['authAttributes'] = $authToken->getDetails();
 
+    $params['given_name'] = $_SERVER['OIDC_CLAIM_given_name'];
```
gregcorbett, line 70: As a note to future Greg, this will need wrapping in a "if using OIDC" statement.

gregcorbett: or rather everything from `$authDetails = $_SERVER['OIDC_CLAIM_external_authn'];` will need to be.

config/gocdb_schema.xml
```diff
@@ -583,7 +583,7 @@
         <field>
             <fname>TYPE</fname>
             <length>255</length>
-            <regex>/^X509$/</regex>
+            <regex>/^(X509|OIDC Subject)$/</regex>
```
gregcorbett, line 586: As a note to Future Greg, at the moment when someone/thing accesses the API via OIDC, we just see a OIDC subject - i.e. we don't see the issuer (we do see the issuer when someone uses the portal.) So I think it makes sense to refer to "OIDC Subject" with regards to API credentials.

I wonder if changing `X509` to `X.509` would break everything for existing credentials...

gregcorbett: > I wonder if changing X509 to X.509 would break everything for existing credentials...

It does, see: https://github.com/gregcorbett/gocdb/tree/X.509

ineilson: I'm not saying there is a problem here related to APIAuthentication, but a note to check, and anyway name consistency might be a good thing

https://github.com/GOCDB/gocdb/blob/58c2c87b606dc2c58bcc022de9c40e807130572a/config/gocdb_schema.xml#L592

gregcorbett: Captured in https://github.com/GOCDB/gocdb/pull/324

htdocs/web_portal/controllers/user/register.php
gregcorbett: Captured in #325

gregcorbett (approved):
🎆

PR #264, opened 2021-01-19
URL-based configuration monitoring fixes

### LINKED ISSUE #263: /portal/GOCDB_monitor/ checks the status of the default server when in a URL-specific view
See https://host-172-16-112-212.nubes.stfc.ac.uk/portal/GOCDB_monitor/

It's checking the status of gocdb-preprod.egi.eu, when it should be checking the status of host-172-16-112-212.nubes.stfc.ac.uk

```
PI URL is: https://gocdb-preprod.egi.eu/gocdbpi/public/?method=get_site_list
Portal URl is: https://gocdb-preprod.egi.eu/portal
Server Base URL is: https://gocdb-preprod.egi.eu
```

ineilson:
Resolves #263 

Enable monitoring pages to use url-based configuration plus minor logic fixes.

### DIFF
--- a/htdocs/web_portal/GOCDB_monitor/check.php
+++ b/htdocs/web_portal/GOCDB_monitor/check.php
@@ -23,7 +23,7 @@ foreach ($res as $r){
 
 if ($counts["error"] != 0) {
     echo("An error has been detected while checking GOCDB services. ".
-            "Please check https://goc.egi.eu/portal/GOCDB_monitor/ to find out more\n");
+        "Please check https://goc.egi.eu/portal/GOCDB_monitor/ to find out more\n");
     exit(2); // return Nagios error code for CRITICAL
 }
 else if ($counts["warn"] != 0) {
--- a/htdocs/web_portal/GOCDB_monitor/check.php
+++ b/htdocs/web_portal/GOCDB_monitor/check.php
@@ -2,9 +2,14 @@
 require_once "tests.php";
 
 $res[1] = test_db_connection();
-$res[2] = test_url(PI_URL);
+$res[2] = test_url(
+            Factory::getConfigService()->GetPiUrl().
+            get_testPiMethod()
+            );
 //$res[3] = test_url(PORTAL_URL);
-$res[3] = test_url(SERVER_BASE_URL);
+$res[3] = test_url(
+            Factory::getConfigService()->getServerBaseUrl()
+            );
 
 
 $counts=array(	"ok" => 0,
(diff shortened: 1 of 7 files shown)


gregcorbett:
Looks good :), could `1e9bdfb` be factored out into it's own separate PR though?

ineilson:
Yes, of course, it was kind of a mistake that it slipped through but I hoped you wouldn't notice ;-)

htdocs/web_portal/GOCDB_monitor/check.php
```diff
@@ -17,7 +22,8 @@
 }
 
 if ($counts["error"] != 0) {
-    echo("An error has been detected while checking GOCDB services. Please check https://goc.egi.eu/portal/GOCDB_monitor/ to find out more\n");
+    echo("An error has been detected while checking GOCDB services. ".
+            "Please check https://goc.egi.eu/portal/GOCDB_monitor/ to find out more\n");
```
gregcorbett, line 26: The indentation feels weird to me, I'd probably prefer something like:

```suggestion
    echo(
        "An error has been detected while checking GOCDB services. ".
        "Please check https://goc.egi.eu/portal/GOCDB_monitor/ to find out more\n"
    );
```

But that's just the python programmer in me speaking, I'm looking into what if anything codeclimate and codacy are actually looking for style wise.

ineilson: As it is is closer to FIG-PSR12 (https://www.php-fig.org/psr/psr-12/) which is the default for vscode, although the second line indent is a bit weird. The whole conditional section auto-formats with psr12 according to vscode as -
```
if ($counts["error"] != 0) {
    echo ("An error has been detected while checking GOCDB services. " .
        "Please check https://goc.egi.eu/portal/GOCDB_monitor/ to find out more\n");
    exit(2); // return Nagios error code for CRITICAL
} else if ($counts["warn"] != 0) {
    echo ("At least one of GOCDB tests shows a warning. It is safe to ignore it anyway\n");
    exit(0); // we don't want notifications if there is just a warning
} else {
    echo ("All GOCDB tests are looking good\n");
    exit(0);
}
```
Your call, php or python?

gregcorbett: We should go for PHP style as it is PHP code. So would that be?:

```suggestion
    echo("An error has been detected while checking GOCDB services. ".
        "Please check https://goc.egi.eu/portal/GOCDB_monitor/ to find out more\n");
```

Vaguely related: I think through some wrangling we can get one of codeclimate/codacy to check specifically for things that go against PSR-12: Extended Coding Style. (Now captured here: https://github.com/GOCDB/gocdb/issues/270)

gregcorbett (approved):
LGTM

PR #275, opened 2021-03-23
Fix adding/editing a Site on MariaDB

### LINKED ISSUE #257: Adding/Editing a Site fails when longitude or latitude are empty on MariaDB
I don't have a MariaDB instance to hand to confirm the exact error, but it was something along the lines of "Invalid float", as if a `0` was getting passed to the database and it wanted `0.0`

ElliottKasoar:
Resolves #257

Currently longitude and latitude are not required and can be left blank, but with MariaDB (and possibly other updated databases) this leads to an exception when adding/editing sites:

> An exception occurred while executing 'UPDATE Sites SET latitude = ?, longitude = ?, alarmEmail = ? WHERE id = ?' with params ["", "", "", 13]: SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect double value: '' for column \`gocdb\`.\`Sites\`.\`latitude\` at row 1

Unentered latitude and/or longitude when adding/editing a site are replaced with `null` to solve this.

### DIFF
--- a/lib/Gocdb_Services/Site.php
+++ b/lib/Gocdb_Services/Site.php
@@ -189,6 +189,10 @@ class Site extends AbstractEntityService{
             throw new \Exception("Invalid scope update action");
         }
 
+        // Replace lat/long with null if no value entered
+        $this->emptyStringToNull($newValues['Site']['LATITUDE']);
+        $this->emptyStringToNull($newValues['Site']['LONGITUDE']);
+
         $this->em->getConnection()->beginTransaction();
         try {
             // Set the site's member variables
@@ -363,6 +367,13 @@ class Site extends AbstractEntityService{
             if (count($errors) > 0) {
                 throw new \Exception($errors[0]); // show the first message.
             }
+        }
+    }
+
+    // Replaces empty string with null
+    private function emptyStringToNull(&$var) {
+        if ($var === '') {
+            $var = null;
         }
     }
 
@@ -727,6 +738,10 @@ class Site extends AbstractEntityService{
 
         //check there are the required number of OPTIONAL scopes specified
         $this->checkNumberOfScopes($values['Scope_ids']);
+
+        // Replace lat/long with null if no value entered
+        $this->emptyStringToNull($values['Site']['LATITUDE']);
+        $this->emptyStringToNull($values['Site']['LONGITUDE']);
 
         // Populate the entity
         try {


tofu-rocketry:
Will the map handle these nicely?

gregcorbett:
Yarp, Sites with both a latitude or longitude of zero don't make it onto the map ([relevant function](https://github.com/GOCDB/gocdb/blob/dev/lib/Gocdb_Services/Site.php#L1303))

GRyall:
Why - what is the benefit from this change?

tofu-rocketry:
> Why - what is the benefit from this change?

Is is the input validation complaining?

ineilson:
A bit of a nit-picky point, but 'empty' has a specific meaning in PHP (https://www.php.net/manual/en/function.empty.php). I don't think that's what the added function emptyToZero is testing for, is it?

ElliottKasoar:
> A bit of a nit-picky point, but 'empty' has a specific meaning in PHP (https://www.php.net/manual/en/function.empty.php). I don't think that's what the added function emptyToZero is testing for, is it?

I wasn't aware of that. The function is testing for `null` or `''`,  which do both evaluate as empty, but perhaps it is misleading as the functionality is more specific. How would you refer to refer to an empty string in PHP?

GRyall:
Looking back at the original issue, I am not sure that fudging the data is the right way to fix this issue. Surely it would be better to understand the error and ensure the field was typed properly. It might even be preferable to ensure a null is passed rather than an empty string if that is what the DB needs.

ElliottKasoar:
I don't have an Oracle implementation, but it had been suggested that the root of the divergent behaviour may be that a more forgiving DB was treating the blank input as 0.0, which I was aiming to encode explicitly.

Also note that #257 may be misleading, as the distinction between integer 0 and double 0.0 does not lead to errors for my DB.

I originally thought I was getting errors for null too, which is why I check for `$var == '' || $var == null`, but actually null can be passed, so an alternative is:

```
private function emptyStringToNull(&$var) {
    if ($var === '') {
        $var = null;
    }
}
```

GRyall:
> 
> 
> I don't have an Oracle implementation, but it had been suggested that the root of the divergent behaviour may be that a more forgiving DB was treating the blank input as 0.0, which I was aiming to encode explicitly.
> 
> Also note that #257 may be misleading, as the distinction between integer 0 and double 0.0 does not lead to errors for my DB.
> 
> I originally thought I was getting errors for null too, which is why I check for `$var == '' || $var == null`, but actually null can be passed, so an alternative is:
> 
> ```
> private function emptyStringToNull(&$var) {
>     if ($var === '') {
>         $var = null;
>     }
> }
> ```

That would seem preferable as it is closer to the intention of what the user is entering and we won't populate our database with a load of zero entries.

GRyall:
Has this been confirmed to fix #257 ? If so, looks good.

ineilson:
> How would you refer to refer to an empty string in PHP?

'zero length string'

ElliottKasoar:
> Has this been confirmed to fix #257 ? If so, looks good.
 It fixes it for me on MariaDB 10.3.20.

> 'zero length string'
Would `zeroLengthStringToNull` be preferable to `emptyStringToNull`? It becomes more convoluted without, to me, adding clarity to the functionality, but I'll happily change it if you disagree (or can suggest a nice alternative).

GRyall (approved):

ElliottKasoar:
(Asked by Greg to change this PR so it merges into the release-5.7.6 branch directly)

gregcorbett (approved):

Issue #272, opened 2021-03-04
Services Groups page not filtered by scope


tofu-rocketry:
See title and nominally IRIS view: https://gocdb.iris.ac.uk/portal/index.php?Page_Type=Service_Groups

PR #278, opened 2021-05-10
Add filtering by default scope to Service Groups view

ineilson:
Partially addresses #272 but scope filtering is inconsistent with implementation for Services - 'any' scope matching is yet to be implemented.


gregcorbett:
Sorry, I thought there was still some work outstanding - with 'all' and 'any' scope matching as per Services display, (hence the convert to draft) but I completely missed the second update :).

gregcorbett:
Could you rebase this to resolve the conflict?

Issue #241, opened 2020-07-20
New feature request: expanded site level logging accessable to users


GRyall:
Following on from https://ggus.eu/index.php?mode=ticket_info&ticket_id=147800, there has been a request for a new feature. IMO the best way to implement this would be to expand the existing role change logging to cover all changes to a site and it's services and endpoints. Thought would need to be given to deleting log entries after a set time and/or limits on log size (especially given increased write API functionality could lead to frequent changes).

gregcorbett:
logging downtime creation and deletion would also be useful.

PR #279, opened 2021-05-10
Prevent downtimes scheduled within 24 hours from being deleted

### LINKED ISSUE #269: Prevent downtimes starting in 24 hours from being deleted.
Highlighted via [GGUS](https://ggus.eu/index.php?mode=ticket_info&ticket_id=149777).

Whilst a `downtime` can't be edited within 24 hours of it's start time "because it has been read in by other systems", it can be deleted. This seems inconsistent.

Checks on what can and can't be done to `downtimes` are split across both the view and controller layer, I would suggest this check is implemented at the controller level.

ElliottKasoar:
Resolves #269

Added a check when attempting to delete a scheduled downtime to prevent deletion if the start time is within 24 hours.

As this check is also carried out for the old and new start times when editing a downtime, this functionality has been factored out into `validateStartTime`.

The additional validation when deleting a downtime has been factored out with the existing check (if the downtime is on-going) for clarity.

### DIFF
--- a/lib/Gocdb_Services/Downtime.php
+++ b/lib/Gocdb_Services/Downtime.php
@@ -523,6 +523,24 @@ class Downtime extends AbstractEntityService{
     }
 
     /**
+     * Check start of scheduled downtime is at least one day in the future
+     * @param $startTime
+     * @param $classification
+     * @throws \Exception if the downtime is not eligible for editing/deleting.
+     */
+    private function validateStartTime($startTime, $classification) {
+        if ($classification === "SCHEDULED") {
+            $nowUtc = new \DateTime(null, new \DateTimeZone('UTC'));
+            $oneDay = \DateInterval::createFromDateString('1 days');
+            $tomorrowUtc = $nowUtc->add($oneDay);
+
+            if($startTime < $tomorrowUtc) {
+                throw new \Exception("Cannot edit or delete a SCHEDULED downtime starting within 24 hours.");
+            }
+        }
+    }
+
+    /**
      * Check with the business rules that the existing downtime's dates
      * allow editing of the downtime.
      * @link https://wiki.egi.eu/wiki/GOCDB/Input_System_User_Documentation#Downtime_shortening_and_extension downtime shortening and extension
--- a/lib/Gocdb_Services/Downtime.php
+++ b/lib/Gocdb_Services/Downtime.php
@@ -550,6 +550,12 @@ class Downtime extends AbstractEntityService{
         if($dt->hasStarted()) {
             throw new \Exception("This downtime has already started.");
         }
+
+        /* @var $startTime \DateTime */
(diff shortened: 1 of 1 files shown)


ElliottKasoar:
(No new changes, I squashed the very minor commits)

ineilson (approved):
Personally, I don't like programs reporting using contractions - "can't" - but use here pre-dates these changes.
Similarly for using short (<2 character) variable names, which angers the 'standard' PHPMD ruleset.
Does this address the issue of logging a failed delete attempt raised in #269?
Otherwise, looks good.

ElliottKasoar:
> Personally, I don't like programs reporting using contractions - "can't" - but use here pre-dates these changes.
> Similarly for using short (<2 character) variable names, which angers the 'standard' PHPMD ruleset.

I can definitely update the exception message, as I've changed it a bit anyway. I did consider changing `$dt`, but it appears 50+ times within Downtime.php, as well as in other scripts, so it would require a lot of changes to maintain consistent naming, which I imagine is a bigger change than we want here?

> Does this address the issue of logging a failed delete attempt raised in #269?

This was discussed briefly with @gregcorbett when I started work on the issue. I think we felt that it made sense to treat the error in the same way as [downtimes that have already started](https://github.com/GOCDB/gocdb/blob/dev/lib/Gocdb_Services/Downtime.php#L665), which wouldn't allow an HTTP 400 to be issued.

(Following discussions with @ineilson) There is (currently) no WriteAPI function to manipulate downtimes, so the above isn't an issue. An alternative would be to instead flag an attempted (authorised) deletion with some indication on the web display, but that would be a more substantial change.

ineilson:
#241 and the referenced GGUS ticket may be relevant here if any future work is planned.

ineilson:
> I can definitely update the exception message, as I've changed it a bit anyway. I did consider changing $dt, but it appears 50+ times within Downtime.php, as well as in other scripts, so it would require a lot of changes to maintain consistent naming, which I imagine is a bigger change than we want here?

The Big Boss @gregcorbett can decide whether 2 char len variable $dt is included in #265, along with $em, $dn and $id.

gregcorbett:
> The Big Boss @gregcorbett can decide whether 2 char len variable $dt is included in #265, along with $em, $dn and $id.

I would say changing `$dt` to something more descriptive throughout the code should be it own separate change in the future.

ElliottKasoar:
> > Personally, I don't like programs reporting using contractions - "can't" - but use here pre-dates these changes.
> 
> I can definitely update the exception message, as I've changed it a bit anyway.

I removed contractions from exceptions/comments where I've made changes, and squashed into existing commits.

ineilson (dismissed):

gregcorbett:
Just rebasing this on top of the latest dev

gregcorbett (approved):

PR #281, opened 2021-05-11
Add non-admin view for scopes

### LINKED ISSUE #273: Need non-admin view of an individual scope
There is an admin-only view of an individual scope (i.e. Page_Type=Admin_Scope&id=X) that lists all the NGIs, Sites, Service Groups and Services in a scope.

Given the rise of scope based views of GOCDB, it would be useful for something similar to be made available to non-admins (without add/edit/delete), as this is the one place in the portal to get an overview of everything in that scope.

I would suggest combining Page_Type=Scope_Help and Page_Type=Admin_Scopes into a single page, and changing Page_Type=Admin_Scope to be able to show or not show delete / edit buttons as appropriate (i.e. based on if the user is an admin or not).

ElliottKasoar:
Resolves #273

As suggested in #273,  Page_Type=Scope_Help and  Page_Type=Admin_Scopes have been combined into a single page,  Page_Type=Scopes, with admin controls inaccessible to non-admin users.

Similarly, the  Page_Type=Admin_Scope page has been replaced with  Page_Type=Scope, which displays the same information, except the admin controls are inaccessible to non-admin users.

Two new controllers, scopes.php and scope.php have also been added, based on the equivalent admin controllers. Their functionality is largely the same, except they pass whether the user is an admin to the views, rather than throwing an exception for non-admin users.

The scopes.php controller also passes lists of reserved and optional scopes, to allow Page_Type=Scopes to display this information in a similar manner to Page_Type=Scope_Help.

The Scope_Help and Admin_Scope views/controllers have been removed. The updated scripts may need their license information etc. updating further?


GRyall (changes requested):
All of those commits need squashing down into self contained commits with commit messages that can be understood without the context of the PR.

GRyall (changes requested):
Could you also confirm you have searched the codebase for instances of Admin_scope and scopes_help?

htdocs/web_portal/index.php
```diff
@@ -187,10 +187,18 @@ function Draw_Page($Page_Type) {
             require_once __DIR__ . '/controllers/project/view_project.php';
             show_project();
             break;
-        case "Scope_Help":
-            //rejectIfNotAuthenticated();
-            require_once __DIR__ . '/controllers/scope_help.php';
-            show_help();
+        // case "Scope_Help":
```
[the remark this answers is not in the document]
ElliottKasoar: I kept them partly to be consistent with `Admin_Change_User_Admin_Status`, which had been commented out. I was also thinking that while it is possible to bring things back, if all references to something have been removed, it becomes increasingly unlikely that someone will know it was there to bring back.

I will remove the commented out lines though. Would you delete the unused scripts as well then?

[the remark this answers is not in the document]
GRyall: Remaining commented out code in GOCDB is likely a legacy from when we used SVN for version control and seeing history was harder. We should only keep code if it is actively used now or we plan to use it in the future.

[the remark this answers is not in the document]
ElliottKasoar: Ah I see, thanks. I've removed the commented out code and the unused scripts now.

ElliottKasoar:
> All of those commits need squashing down into self contained commits with commit messages that can be understood without the context of the PR.

Is it ok now?

> Could you also confirm you have searched the codebase for instances of Admin_scope and scopes_help?

Yes I have. There were still some references in the unused/undeleted scripts, but I can't see any now I've deleted them too.

htdocs/web_portal/index.php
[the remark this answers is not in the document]
GRyall: > 
> > All of those commits need squashing down into self contained commits with commit messages that can be understood without the context of the PR.
> 
> Is it ok now?
> 

Yep!

GRyall:
> 
> 
> > All of those commits need squashing down into self contained commits with commit messages that can be understood without the context of the PR.
> 
> Is it ok now?
> 


Yep!

GRyall (commented):
I don't see any issues, but I don't currently have a set-up to test the changes against - so I will leave the final approval to someone who does.

ineilson (approved):

htdocs/web_portal/controllers/scopes.php
```diff
@@ -1,13 +1,13 @@
...
- * Author: George Ryall, David Meredith
- * Description: Retrieves and draws the data for a user
+ * File: view_service_types.php
+ * Author: George Ryall, David Meredith, Elliott Kasoar
+ * Description: Controller for showing all scope tags available in GOCDB
  *
  * License information
  *
- * Copyright 2013 STFC
+ * Copyright � 2013 STFC
```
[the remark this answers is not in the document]
ElliottKasoar: Ah interesting, it comes from [here](https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/controllers/admin/scopes.php#L10). I basically didn't touch the license info, but it's decided that the final file is closer to scope_help.php, so it looks like I added it. I'll remove it at the same time as your other suggestion.

Should the copyright be updated to 2021, or any other changes be made?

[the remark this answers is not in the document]
ineilson: > Should the copyright be updated to 2021, or any other changes be made?

I've no clue on this. I would say yes to the copyright but, in truth, doc comments seem largely aspirational and not enforced, so that's a question for @gregcorbett I think.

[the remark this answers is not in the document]
tofu-rocketry: In theory, you should add each year that a release is made, collapsing them into a range where appropriate. e.g. releases in 2001, 2002, 2003, 2005 gives 2001-2003, 2005

This is something that should probably be done at release time.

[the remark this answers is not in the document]
gregcorbett: We should at least get rid of the non-displayable character here.

htdocs/web_portal/views/scopes.php
```diff
@@ -1,6 +1,75 @@
...
+                <a href="index.php?Page_Type=Admin_Add_Scope">
+                <img src="<?php echo \GocContextPath::getPath()?>img/add.png" height="25px" />
+                <br />
+                <span>Add Scope</span>
+                </a>
+            </center>
+        </div>
+    <?php endif; ?>
+
+    <?php $numberOfScopes = sizeof($params['Scopes'])?>
```
ineilson, line 24: I rather agre with the comments at https://www.php.net/manual/en/function.sizeof.php and prefer count() instead of sizeof();
```suggestion
    <?php $numberOfScopes = count($params['Scopes'])?>
```

htdocs/web_portal/views/scopes.php
```diff
@@ -1,6 +1,75 @@
...
+                <th class="site_table">Name</th>
+                <th class="site_table">Description</th>
+                <th class="site_table">Reserved?</th>
+                <?php if(!$params['portalIsReadOnly'] && $params['UserIsAdmin']):?>
+                    <th class="site_table" style="width: 10%">Remove</th>
+                <?php endif; ?>
+            </tr>
+            <?php
+            $num = 2;
+            if(sizeof($numberOfScopes > 0)) {
```
ineilson, line 40: Surely ?
```suggestion
            if($numberOfScopes > 0) {
```

ineilson (changes requested):

htdocs/web_portal/views/scopes.php
ElliottKasoar: sizeof() is used in quite a few places throughout the code, so it might be worth fixing them all at some point. I'll change it everywhere it appears in scopes.php and scope.php for now though.

htdocs/web_portal/views/scopes.php
ElliottKasoar: I would think so, but does that line actually do anything (I didn't write it)? Doesn't the foreach that follows mean that if there are no scopes, no new HTML is added anyway?

ineilson: Yes, looks like it. Imho having it improves readability but it might be better put to use to stop the table header being displayed when there is no content. But that might be more of a change than is wanted here. Minimum I'd say is the 'if' clause should be corrected (even if it always resolves to true, or deleted.

htdocs/web_portal/views/scopes.php
```diff
@@ -1,6 +1,75 @@
...
+                <span>Add Scope</span>
+                </a>
+            </center>
+        </div>
+    <?php endif; ?>
+
+    <?php $numberOfScopes = count($params['Scopes'])?>
+    <div class="listContainer">
+        <span class="header listHeader">
+            <?php echo $numberOfScopes ?> Scope<?php if($numberOfScopes) echo "s"?>
```
ineilson, line 27:
```suggestion
            <?php echo $numberOfScopes ?> Scope<?php if ($numberOfScopes  != 1) echo "s"?>
```

ElliottKasoar: Yep, but is `!==` not generally preferable?

ineilson (approved):
Finger trouble reviewing. Approved subject to  https://github.com/GOCDB/gocdb/pull/281/files?file-filters%5B%5D=.php&file-filters%5B%5D=.xml&hide-deleted-files=true#r672099274

htdocs/web_portal/views/scopes.php
ineilson: I guess we could write a book on that one 😀 It's not an operator I'm in the habit of using but you're probably correct ... generally.

ElliottKasoar:
(Squashed requested change into last commit)

ineilson (approved):

htdocs/web_portal/views/scopes.php
gregcorbett: `==` is loose comparison, `===` is strict comparisons. For simple types (int, string, bool) I would use strict comparisons because it seems you can get into a situation where  `true` is loosely `false` at least under older versions of php.

https://www.php.net/manual/en/types.comparisons.php
https://news.ycombinator.com/item?id=9485855

ElliottKasoar:
Might it be worth hiding some of the information behind a help button, similar to the Downtimes Calendar page?

gregcorbett:
>Might it be worth hiding some of the information behind a help button, similar to the Downtimes Calendar page?

What information did you have in mind?

ElliottKasoar:
> What information did you have in mind?

I was thinking potentially some/all of the sections from the Scope_Help page that weren't on the Admin_Scopes page: `What are scope tags?`, `What are Reserved tags?`, `How are scope tags used in the API?` and `What does having the EGI scope applied mean?`.

I would imagine most people would visit the page for the list of scopes itself, similar to most of the other pages, which was partly why my combined page has the table first. The explanatory information could end up quite far down the page, particularly as the number of scopes increases, which isn't necessarily as convenient as having it hidden but displayable at the top. I'm not sure though, it was just an idea.

gregcorbett:
That sounds like a good improvement to make 😄

ElliottKasoar:
I would have preferred the help button to be styled more consistently with the downtimes calendar help button, but it didn't seem to fit as well, particularly with the Add Scope button, although most people won't see that. I can change it back if it people think it would look better.

gregcorbett:
perhaps we could have some kind of [Collapsible](https://www.w3schools.com/howto/howto_js_collapsible.asp) if it could be style to fit in with the rest of GOCDB?

ineilson:
The Calendar page 'header' anyway seems inconsistent with other pages in missing a one-line description. Perhaps such a line with a collapsible 'Show more ...', or similar,  would work consistently for both cases. The help and question-mark does look out of place IMHO.

ElliottKasoar:
> perhaps we could have some kind of [Collapsible](https://www.w3schools.com/howto/howto_js_collapsible.asp) if it could be style to fit in with the rest of GOCDB?

Hmm it feels like a collapsible might be a little intrusive, but perhaps it is still the best solution.

> The Calendar page 'header' anyway seems inconsistent with other pages in missing a one-line description. Perhaps such a line with a collapsible 'Show more ...', or similar, would work consistently for both cases. The help and question-mark does look out of place IMHO.

To clarify, are you suggesting a collapsible in the same place as other one-lines, or add a similar one-line as well as a collapsible below (e.g. similar to the Projects and NGIs pages, but a collapsible rather than a "What is a..." link?)

In terms of general consistency, the Scopes page is also lacking own icon.

ineilson:
> To clarify, are you suggesting a collapsible in the same place as other one-lines, or add a similar one-line as well as a collapsible below (e.g. similar to the Projects and NGIs pages, but a collapsible rather than a "What is a..." link?)

The latter: one liner plus collapsible. 

> In terms of general consistency, the Scopes page is also lacking own icon.

A nice hackathon/first change issue for somebody then. 😃

ElliottKasoar:
As discussed with @ineilson and @gregcorbett, toggling to show/hide the information is now done via a link, the text of which changes to provide the correct context.

ineilson (approved):

gregcorbett:
rebasing on the latest dev

gregcorbett (approved):

PR #283, opened 2021-06-21
Pre privacy updates

ineilson:
Miscellaneous, non-privacy-related changes, in preparation for data privacy updates.

PR #284, opened 2021-06-21
Personal data privacy 

ineilson:
Depends on #283 

Implement restrictions aligning exposure of personal data in GOCDB with requirements of the EU General Data Protection Regulation (GDPR) and UK GDPR.
Simply: if the local_info.xml configuration restrict_personal_data is 'true', only users with a role at a site, or API queries authenticated with a registered APIAuthentication credential, can view personal data.  A 'false' value honours the current behaviour, by which 'only' a registered IGTF certificate credential is required.


htdocs/PI/index.php
```diff
@@ -367,11 +370,48 @@ function getXml() {
...
+        $authenticated = false;
+
+        $user = \Factory::getUserService()->getUserByPrinciple($this->dn);
+
+        if ($user == null) {
+            // Incoming credential is not that of a registered user.
+            // Check if it is registered API Authentication credential.
+
+            $authEntServ = \Factory::getAPIAuthenticationService();
+            $authEnt = $authEntServ->getAPIAuthentication($this->dn, "X509");
```
gregcorbett, line 393: with `"X509"` being hard coded here, what happens if `$this->dn` isn't a DN? What if it's a EGI Check-In identifier?

ineilson: The honest answer is that I don't know, but since X509 is hard coded in the available types on instance creation, it seems the safest option for now. Has the identification of an EGI identifier progressed more outside of the dev branch? https://github.com/GOCDB/gocdb/blob/facc3167e2b28be0524ddeaa3241edbdcfe12008/htdocs/web_portal/controllers/site/add_api_auth.php#L61

gregcorbett: > Has the identification of an EGI identifier progressed more outside of the dev branch?

Sarah's work will add "OIDC Subject", so I think there will be some interplay here. Hopefully not too much as the `APIAuthenticationService` looks pretty "type" agnostic to me :).

ineilson: Let's hope so. There doesn't seem much that can be done regarding this PR?

ineilson: Out of interest: why 'OIDC Subject' rather than just 'OIDC', which seems more in line with 'X509' rather than 'X509 Subject'? Do we expect 'OIDC Somethingelse'?

gregcorbett: > Let's hope so. There doesn't seem much that can be done regarding this PR?

No yet 😄 

> Out of interest: why 'OIDC Subject' rather than just 'OIDC', which seems more in line with 'X509' rather than 'X509 Subject'? Do we expect 'OIDC Somethingelse'?

We should probably take this over to #260 😋

gregcorbett:
Could you rebase this and resolve the conflicts?

gregcorbett:
I note that if restricting personal data is in effect, you can only see personal data if you have a Site level role - not a NGI or Project level role.

Could this be extended so that having a NGI or Project level role is sufficent?

Given #364, I agree that currently a ServiceGroup role should not be sufficent to see personal data.

gregcorbett:
The NGI level contact info and user / role info should also be hidden if one doesn't have a role.

gregcorbett:
If you don't have a role, you can't see the "Request Role" button on a Site or Project entity

image

So how can you request a role? We still need the request roll button to be visible.

gregcorbett:
There's also a nuance whereby:
- If I have an NGI role, I can add an API identifier to a Site (so far so good)
- I may later gain a role over that site (okay...)
- I can't have that site role revoked while the API identifier is in place, even though I'd still have the NGI role (booo)

gregcorbett:
I've captured the follow on work identified above in #366. Given that the current production behaviour is preserved if `restrict_personal_data` is set to `false`, I'm going to merge this.

PR #287, opened 2021-06-29
Fix QueryPopulateSSO_UsernameRunner

ElliottKasoar:
QueryPopulateSSO_UsernameRunner.php has not been working for some time, as the EGI URL (`$url`) has changed, and the proxy does not seem to work. This results in `$username1` not being populated or updated when the script is run.

`$url` has therefore been updated and the proxy settings have been removed. This appears to fix the script, correctly populating `$username1`.

### DIFF
--- a/resources/QueryPopulateSSO_UsernameRunner.php
+++ b/resources/QueryPopulateSSO_UsernameRunner.php
@@ -39,12 +39,10 @@ foreach ($users as $user) {
     ++$count;
     $cleanDN = cleanDN($user->getCertificateDn());
     if (!empty($cleanDN)) {
-        $url = "https://www.egi.eu/sso/api/user?dn=" . $cleanDN;
+        $url = "https://sso.egi.eu/admin/api/user?dn=" . $cleanDN;
         $ch = curl_init();
         curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false);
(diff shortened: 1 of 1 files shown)


resources/QueryPopulateSSO_UsernameRunner.php
```diff
@@ -39,12 +39,10 @@
     ++$count;
     $cleanDN = cleanDN($user->getCertificateDn());
     if (!empty($cleanDN)) {
-        $url = "https://www.egi.eu/sso/api/user?dn=" . $cleanDN;
+        $url = "https://sso.egi.eu/admin/api/user?dn=" . $cleanDN;
         $ch = curl_init();
         curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false);
```
ineilson, line 44: Not a problem introduced by this PR.

Maybe I'm misreading the curl manual, but for SSL_VERIFY_HOST -

image

 it does imply to me that we are not checking that we trust the issuer CA of the server's certificate.

```suggestion
        curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, true);
```

gregcorbett: Captured #333

gregcorbett (approved):

PR #288, opened 2021-07-01
 Create GitHub Action scripts to replicate travis build testing

ElliottKasoar:
As the travis build is no longer working, this aims to replicate the testing via GitHub Actions.

continuous-integration.yml performs a similar role to .travis.yml, setting up jobs to test different combinations of php, databases, etc. Different extensions, operating systems, php/database versions could be added straightforwardly via the matrices. However, testing of different databases may require separate jobs, as is done for SQLite, as it appears `services` cannot be conditional.

The scripts from .travis have largely been copied to .github/actions, with minor changes for compatibility with the different directory structure, variables and database setup.


tofu-rocketry:
@gregcorbett Did we want to re-enable the Travis build on the .com site for comparison? It's not too difficult (though a little fiddly in places).

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,146 @@
+name: "Continuous Integration"
+
+on: [push, pull_request]
+
+jobs:
+
+  phpunit-sqlite:
+    name: "PHPUnit with SQLite"
```
tofu-rocketry, line 8: The names seem to have versions appended automatically, right? e.g. this one comes out as "PHPUnit with SQLite (5.6)"

Is there any way to control that? For the MariaDB tests, it's not clear which number refers to which thing.

.github/actions/run_tests.sh
```diff
@@ -1,14 +1,12 @@
 #!/bin/bash
 
-if [[ $TRAVIS ]]; then
-    cd "$TRAVIS_BUILD_DIR" || exit 2
-
+if [[ $DIR ]]; then
```
tofu-rocketry, line 3: Where's `DIR` come from?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,146 @@
...
+            mariadb-version: "10.3"
+            extension: "pdo_mysql"
+            experimental: true
+          - php-version: "8.0"
+            mariadb-version: "10.3"
+            extension: "pdo_mysql"
+            experimental: true
+
+    services:
+      mariadb:
```
tofu-rocketry, line 95: How'd you come up with this service?

FAO: @pjcon

.github/actions/run_tests.sh
ElliottKasoar: It's passed from continuous-integration.yml via the steps that call run_test.sh:
```
env:
  DIR: ${{ github.workspace }}
```
I think it's more of a replacement for TRAVIS_BUILD_DIR (`DIR: /home/runner/work/gocdb/gocdb`), but as it should only be set when the action is run, I think it serves the same purpose.

.github/workflows/continuous-integration.yml
ElliottKasoar: I think I originally saw it being used by Doctrine:
https://github.com/doctrine/dbal/blob/3.1.x/.github/workflows/continuous-integration.yml#L224

Some of the options came from the Docker documentation (but probably elsewhere too):
https://hub.docker.com/_/mariadb/

.github/workflows/continuous-integration.yml
ElliottKasoar: Yes, it's automatic by default. It seems that if you include ${{ matrix.X}} values in the name, you can overwrite this,  e.g.

`name: "PHPUnit with MariaDB (${{ matrix.mariadb-version }}). PHP: ${{ matrix.php-version }}, extension: ${{ matrix.extension }}, experimental: ${{ matrix.experimental }}"`

gives "PHPUnit with MariaDB (10.3). PHP: 5.4, extension: pdo_mysql, experimental: false" etc.

(https://github.com/ElliottKasoar/gocdb/actions/runs/990052868)

The names don't need to include every matrix value, or even be unique, e.g.

`name: "PHPUnit with MariaDB (${{ matrix.mariadb-version }})"`

gives "PHPUnit with MariaDB (10.3)" for all of them.

(https://github.com/ElliottKasoar/gocdb/runs/2962602044)

Should I commit something like the former (possibly without experimental?)?

gregcorbett:
> @gregcorbett Did we want to re-enable the Travis build on the .com site for comparison?

I do not 😛

.github/workflows/continuous-integration.yml
gregcorbett: >  (possibly without experimental?)

Yes I don't think we need that.

ElliottKasoar: Would you include the SQLite version even though it's not explicitly set?

gregcorbett: How does SQLite get "installed" (for lack of a better term) in the environment the test runs in?

Is it simply there in the `ubuntu-18.04` VM? Does doctrine install it?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,146 @@
...
+        env:
+          DIR: ${{ github.workspace }}
+
+      - name: "Upload to Codecov"
+        uses: "codecov/codecov-action@v1"
+
+
+  phpunit-mariadb:
+    name: "PHPUnit with MariaDB"
+    runs-on: ubuntu-18.04
```
gregcorbett, line 60:
```suggestion
    runs-on: ubuntu-latest
```

Does this work? or do we need an older version of ubuntu for PHP5.X?

.github/actions/run_tests.sh
tofu-rocketry: Ok, then I'd recommend a more specific and descriptive name.

.github/workflows/continuous-integration.yml
ElliottKasoar: I actually had that originally and I think it worked fine. I thought it might be better to specify a version, as I assume "latest" will change over time, but if that's not a problem I'll (check and) change it back.

.github/workflows/continuous-integration.yml
ElliottKasoar: It seems to come pre-installed:
https://github.com/actions/virtual-environments/blob/ubuntu18/20210628.1/images/linux/Ubuntu1804-README.md

.github/workflows/continuous-integration.yml
gregcorbett: > I assume "latest" will change over time

I will, and I think we should only specify a version if, in the future, running on latest causes a problem.

.github/workflows/continuous-integration.yml
gregcorbett: Then maybe just a comment in the yml file pointing people to the [documentation](https://github.com/actions/virtual-environments/tree/ubuntu18/20210628.1/images/linux) is sufficient. Unless it can be figured out programmatically, putting any version number for SQLite in the job name is just going to get out of date.

I do agree that something like the below is easier to read than the default.

`name: "PHPUnit with MariaDB: ${{ matrix.mariadb-version }}, PHP: ${{ matrix.php-version }}, extension: ${{ matrix.extension }}."`

.github/actions/run_tests.sh
ElliottKasoar: Although functionally it'll be the same, I've realised that it would be more appropriate to use $GITHUB_ACTIONS, as this seems to be the direct equivalent of $TRAVIS - it's a default environmental variable, set to true when GitHub Actions is running the workflow.

Given the `else` output largely explaining its function, would simply replacing $TRAVIS  with $GITHUB_ACTIONS be clear enough?

tofu-rocketry: Yup, that's better.

.github/actions/mysql_bootstrap_doctrine.php
```diff
@@ -14,11 +14,11 @@
...
 	'driver' => 'pdo_mysql',
-	'user' => 'travis',
-	'password' => '',
-	'host' => 'localhost',
+	'user' => 'user',
+	'password' => 'password',
+	'host' => '172.18.0.1',
 	'dbname' => 'doctrine',
-    'charset' => 'UTF8'
+	'charset' => 'UTF8'
```
tofu-rocketry, line 21: This uses tabs whereas the other PHP files use spaces.

gregcorbett (approved):
The checks did run (and the important ones pass), one can see them [here](https://github.com/ElliottKasoar/gocdb/actions/runs/1848355522).

I guess they aren't showing up here because the github actions aren't part of the dev yet.

gregcorbett (approved):

PR #294, opened 2021-08-06
Implement identity linking

ElliottKasoar:
A new entity, UserProperty, implemented similarly to other extension properties (e.g. [SiteProperty ](https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/entities/SiteProperty.php) etc.), stores key-value pairs associated with users, representing the authentication type and ID string for identifiers. This allows each user to have multiple associated identifiers, rather than the single ID string stored in certificateDn, and provides flexibility, as authentication mechanisms can be added or removed without further changes to the schema.

The account retrieval [service](https://github.com/GOCDB/gocdb/blob/dev/lib/Gocdb_Services/RetrieveAccount.php) and [entity](https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/entities/RetrieveAccountRequest.php) have been updated to a combined identity linking and account recovery service and entity. The choice of process is determined by whether the user selects authentication type they are currently using (recovery) or a different authentication type (linking).

As previously, the mechanism involves generation of a random code, forming part of a validation URL, which is contained in an email sent to the account associated with the details entered. However, registered users can now carry out the recovery/linking process, involving an attempt to "merge" their roles, before their second account is deleted. An email is also sent to both accounts, if the two have different emails. The creation time of requests is also stored, allowing a script to delete requests after a certain period of time.

Admin and non-admin views of users have been updated to show all ID string associated with a user, as well as allow admin editing of user properties and removal of properties.

A "preferred" ID string is also shown on user pages, in logs and in API outputs. The preference is determined by the order of tokens in MyConfig1, as well as the hardcoded order of authentication realms within each token if it has multiple realms. The API output can alternatively be set to list multiple ID strings, depending on configuration settings.

Scripts to deploy and test GOCDB have also been updated to make use of user properties.

References to `$dn` etc. remain throughout the code, as updating them here would obscure the logic changes.


GRyall:
There is some significant squashing of commits needed here.

I am not sure about userProperty as the new entity, do you really envisage it being as generic as the site Key/value properties? I wonder if there would be value in making clear the entity relates to identity management.

gregcorbett:
> I am not sure about userProperty as the new entity, do you really envisage it being as generic as the site Key/value properties? I wonder if there would be value in making clear the entity relates to identity management.

Hmmmm.... yes... it probably would be better to have separate entities for UserIdentifiers and any future UserProperties we might want to add (even if both entities extend the same base class that does the "heavy lifting" of wrapping a key value pair into a class.

GRyall:
> 
> 
> > I am not sure about userProperty as the new entity, do you really envisage it being as generic as the site Key/value properties? I wonder if there would be value in making clear the entity relates to identity management.
> 
> Hmmmm.... yes... it probably would be better to have separate entities for UserIdentifiers and any future UserProperties we might want to add (even if both entities extend the same base class that does the "heavy lifting" of wrapping a key value pair into a class.

It just seems like a bit of a security risk storing user credentials and arbitrary properties in the same place

ElliottKasoar:
> There is some significant squashing of commits needed here.

There were originally ~200. I'm happy to squash more, but @gregcorbett asked me to preserve a lot of the history, including keeping a few specific commits with very minor changes to the code isolated. I'm sure there's a middle ground, but a lot of the obviously squashable commits are deliberate.

> I am not sure about userProperty as the new entity, do you really envisage it being as generic as the site Key/value properties? I wonder if there would be value in making clear the entity relates to identity management.

> Hmmmm.... yes... it probably would be better to have separate entities for UserIdentifiers and any future UserProperties we might want to add (even if both entities extend the same base class that does the "heavy lifting" of wrapping a key value pair into a class.

My expectation was that the entity's use would remain specific to identity management, which I agree should be clear, but I thought there was also value in recognising the significant structural similarity to other extension properties. If there is a chance that more generic user properties will be added, I agree it doesn't make sense to store them together, and something like UserIdentifiers would definitely make more sense.

lib/Authentication/AuthTokens/X509AuthenticationToken.php
```diff
@@ -26,7 +26,7 @@ public function __construct() {
         //$this->logger->pushHandler(new StreamHandler(__DIR__.'/../../../gocdb.log', Logger::DEBUG));
 
         $this->initialDN = $this->getDN();
-        $this->userDetails = array('AuthenticationRealm' => array('IGTF'));
+        $this->userDetails = array('AuthenticationRealm' => array('IGTF X509 Cert'));
```
ineilson, line 29: Is there anything in the program that enforces the X509 'flavour' to be IGTF only? What set of certificate issuers is accepted is set by the Apache/server configuration I think. Would plain 'X509' be more understandable/general?

gregcorbett: I think I may have originally suggested `IGTF X509 Cert`, but you are right @ineilson, there is nothing in the codebase that enforces the 'flavour' that's all handled by apache.

`X.509` _is_ more future proof (though I can't see a requirement for non IGTF certificate based access appearing anytime soon).

htdocs/web_portal/components/Get_User_Principle.php
```diff
@@ -158,11 +158,23 @@ function Get_User_Principle(){
...
         // once for the current request.
         if ($user !== null) {
             $serv->updateLastLoginTime($user);
+
+            // If property for current auth does not exist, add to user
+            if (!$authExists) {
+                // Get type of auth logged in with e.g. IGTF)
+                $authType = $auth->getDetails()['AuthenticationRealm'][0];
+                $propArr = array($authType, $principleString);
+                $serv->addUserProperty($user, $propArr, $user);
```
gregcorbett: After setting up the User Property, shouldn't we also set the certificate DN to some unique placeholder value?

ElliottKasoar: There aren't immediate problems with having the ID string stored in both places, which is why at this point I hadn't yet overwritten the certificate DN. When it is overwritten, it's done within `addUserProperty`, so this section of the code wouldn't actually change.

I implemented the overwrite from the [registration](https://github.com/GOCDB/gocdb/pull/294/commits/6d03e14a073bc439b1a671789427aa235f3cfc68) onwards, as that was the point when setting the certificate DN was actually necessary, although the final form requires the [schema change](https://github.com/GOCDB/gocdb/pull/294/commits/3c75b41c5ef9a41adb80aa321adbc26bb4a48f92) (which is a separate commit, as you requested).

gregcorbett: It would be clearer if the overwriting of the old certificateDN field was factored out into it's own method and called here.

lib/Gocdb_Services/User.php
```diff
@@ -514,6 +521,9 @@ protected function addUserPropertyLogic(\User $user, array $propArr) {
 
         // $this->addUserPropertyValidation($keyName, $keyValue);
 
+        // Set certificate DN to unique ID for now
+        $user->setCertificateDn($user->getId());
```
gregcorbett: This is a good candidate to squash away, as it later changes to `$user->setCertificateDn(null);`

ElliottKasoar: Do you still want the rest of the changes to registration as a separate commit, or can I squash that into the login changes?

gregcorbett: > Do you still want the rest of the changes to registration as a separate commit,

Yes please.

lib/Gocdb_Services/LinkIdentity.php
```diff
@@ -0,0 +1,420 @@
...
+        . "\n\nRequested new identifier " . ($isLinking ? "to be added to your account" : "with updated ID string") . ":"
+        . "\n\n    - Authentication type: $currentAuthType"
+        . "\n    - ID string: $currentIdString";
+
+        if ($isRegistered) {
+            $body .= "\n\nThis new identifier is currently associated with a second registered account."
+            . " If " . ($isLinking ? "identity linking" : "account recovery") . " is successful, any roles currently associated with this second account ($currentIdString)"
+            . " will be requested for your primary GOCDB account ($primaryIdString)."
+            . " These roles will be approved automatically if either account has permission to do so."
+            . "\n\n Your second account will then be deleted.";
```
gregcorbett, line 262:
```suggestion
            . "\n\nYour second account will then be deleted.";
```

htdocs/web_portal/components/Get_User_Principle.php
ElliottKasoar: Thinking about this more, calling an overwrite method here would come with downsides to consider.

The first is hopefully fairly minor: the same function would have to be called in other parts of the code (e.g. during linking/recovery with an old account), which increases the risk that it isn't called when needed. This isn't really a reason against doing it, more something to be aware/careful of.

More significantly, calling it here would mean that overwriting certificateDn and adding the user property would not be wrapped in a single transaction, unless we added that here too. Overwriting could be done after the user property has been successfully added, which would be less risky, but then validation that the ID string is unique would be made more complicated, as it checks against certificateDns as well as other user properties.

gregcorbett: So we definitely want overwriting `certificateDn` and adding the User Identity wrapped in a single transaction, as we don't want a User to get into a state where they have both a valid `certificateDn` entry and a User Identity (or worse they have neither of those things)

maybe we need a `migrateUserCredentials` method (for example) for this case to wrap the logic of `addUserPropertyLogic` and "overwrite certificateDn" into a single transaction.

lib/Authentication/AuthTokens/X509AuthenticationToken.php
gregcorbett: This was done.

htdocs/web_portal/components/Get_User_Principle.php
gregcorbett: This was done.

lib/Gocdb_Services/PI/QueryBuilders/ParameterBuilder.php
```diff
@@ -143,12 +143,28 @@ private function ifSet($parameters, $qb) {
     }
 
     if (isset($parameters ['dn'])) {
-        $qb->andWhere($qb->expr()->like('u.certificateDn', '?' . ++$bc));
+        ++$bc;
+        $qb->leftJoin('u.userIdentifiers', 'up');
+        $qb->andWhere($qb->expr()->orX(
+            $qb->expr()->andX(
+                $qb->expr()->like('up.keyValue', '?' . $bc),
+                $qb->expr()->like('up.keyName', $qb->expr()->literal('X.509'))
```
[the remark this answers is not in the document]
gregcorbett: resolved in [899984f](https://github.com/GOCDB/gocdb/pull/294/commits/899984f0528a9183163f1383305ee0013328cbe5)

lib/Gocdb_Services/PI/QueryBuilders/ParameterBuilder.php
```diff
@@ -143,12 +143,28 @@ private function ifSet($parameters, $qb) {
...
     }
 
     if (isset($parameters ['dnlike'])) {
-        $qb->andWhere($qb->expr()->like('u.certificateDn', '?' . ++$bc));
+        ++$bc;
+        $qb->leftJoin('u.userIdentifiers', 'up');
+        $qb->andWhere($qb->expr()->orX(
+            $qb->expr()->andX(
+                $qb->expr()->like('up.keyValue', '?' . $bc),
+                $qb->expr()->like('up.keyName', $qb->expr()->literal('X.509'))
```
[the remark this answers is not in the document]
gregcorbett: [899984f](https://github.com/GOCDB/gocdb/pull/294/commits/899984f0528a9183163f1383305ee0013328cbe5)

lib/Gocdb_Services/User.php
```diff
@@ -422,31 +422,479 @@ public function deleteUser(\User $user, \User $currentUser = null) {
...
+
+        // For each auth type, check if an identifier matches
+        foreach ($identifiers as $identifier) {
+            if ($identifier->getKeyName() === $authType) {
+                $idString = $identifier->getKeyValue();
+            }
+        }
+
+        // If no user identifiers and want X.509, return certificateDn
+        if (count($identifiers) === 0 && $authType === 'X.509') {
```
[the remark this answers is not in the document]
gregcorbett: Actually, I think this does preserve existing behavior.

Looking at where this method is called, for users who don't have any user identifiers yet, we **want** to return the the old certDN field - because it's the only identifier with have for them (and it is probably a certificate dn).

lib/Gocdb_Services/LinkIdentity.php
```diff
@@ -0,0 +1,417 @@
...
+        $link = $portalUrl."/index.php?Page_Type=User_Validate_Identity_Link&c=" . $code;
+
+        // Compose email to send to primary user
+        $isPrimary = true;
+        $composedPrimaryEmail = $this->composeEmail($primaryIdString, $currentIdString, $primaryAuthType, $currentAuthType, $isLinking, $isRegistered, $isPrimary, $link);
+        $primarySubject = $composedPrimaryEmail['subject'];
+        $primaryBody = $composedPrimaryEmail['body'];
+
+        // If "sendmail_from" is set in php.ini, use second line ($headers = '';):
+        $headers = "From: no-reply@goc.egi.eu";
```
gregcorbett, line 307: This should be 
```suggestion
        $headers = "From: GOCDB <gocdb-admins@mailman.egi.eu>";
```

gregcorbett: see 126af81

gregcorbett (approved):

PR #295, opened 2021-08-06
Define character set and collation for entity tables

ElliottKasoar:
Define all entity `@Table` character sets and collations, with `"collate"="utf8mb4_bin"` and `"charset"="utf8mb4"`, to ensure current Oracle behaviour (case sensitivity and full character support) is reflected in MariaDB.

This should not change the schema for existing databases, as Doctrine does not appear to update existing tables if this is added, but it will change the SQL generated when creating new tables for MariaDB. When creating tables for Oracle, these changes do not appear to make a difference.

Note: these options cannot be defined for `@JoinTable`s, such as [NGIs_Scopes](https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/entities/NGI.php#L69), so these will have the default character set (utf8) and collation (utf8_unicode_ci), but as these tables only store IDs, this should not be a problem.

### DIFF
--- a/lib/Doctrine/entities/APIAuthentication.php
+++ b/lib/Doctrine/entities/APIAuthentication.php
@@ -20,7 +20,7 @@
   *
   * @author George Ryall (github.com/GRyall)
   *
-  * @Entity @Table(name="APIAuthenticationEntities",uniqueConstraints={@UniqueConstraint(name="siteIdentifier", columns={"parentSite_id", "type", "identifier"})})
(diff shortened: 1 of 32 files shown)


ineilson:
What is the rationale behind choosing the `utf8mb4_bin` collation rather than, say, the default `utf8mb4_general_ci` ? Ref:https://mariadb.com/kb/en/supported-character-sets-and-collations/

ElliottKasoar:
> What is the rationale behind choosing the utf8mb4_bin collation rather than, say, the default utf8mb4_general_ci ? > > Ref:https://mariadb.com/kb/en/supported-character-sets-and-collations/

`utf8mb4_general_ci` is case insensitive, unlike Oracle's default behaviour. The collation needs to match the charset (`utf8mb4_`), and `bin` gives the case sensitive behaviour.

ineilson (approved):
Seems good, with the proviso that I can't think why we're using case sensitivity in the first place, but that's probably another issue.

PR #298, opened 2021-08-20
Sort API output

ElliottKasoar:
All top level entities are sorted by ID using the query builder, apart from countries which are sorted alphabetically, ignoring case. Entities accessed via entity->getX are sorted by their ID using Doctrine Criteria through a helper function.

This provides consistency, which is particularly useful for comparing different databases' API outputs (e.g. Oracle and MariaDB).

### DIFF
--- a/lib/Gocdb_Services/PI/GetDowntime.php
+++ b/lib/Gocdb_Services/PI/GetDowntime.php
@@ -23,6 +23,7 @@ require_once __DIR__ . '/IPIQueryRenderable.php';
 
 
 use Doctrine\ORM\Tools\Pagination\Paginator;
+use Doctrine\Common\Collections\ArrayCollection;
 
 /**
  * Return an XML document that encodes the downtimes with optional cursor-based paging.
@@ -168,7 +169,6 @@ class GetDowntime implements IPIQuery, IPIQueryPageable, IPIQueryRenderable
             ->join('s.ngi', 'n')
             ->join('s.country', 'c')
             ->join('se.serviceType', 'st')
-            //->orderBy('d.startDate', 'DESC')
             ; //try just joins!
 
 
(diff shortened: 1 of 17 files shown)


GRyall:
Before merging this into prod, we should announce it to the community as it changes the behaviour of the API

gregcorbett:
@ElliottKasoar: I presume this doesn't enforce the current "ordering"? But rather a sane alternative that will be consistent across DB technologies?

lib/Gocdb_Services/PI/GetDowntime.php
```diff
@@ -863,11 +882,17 @@ private function getXMLNestedNoPage($downtimes)
                 if ($this->renderMultipleEndpoints) {
                     // Slice the service's endpointLocations array and store just each endpointLocation's Id
                     $currentServiceEndpointIDs = array();
+
+                    // No benefit sorting
```
gregcorbett, line 886: No benefit sorting?

gregcorbett:
Ah, so using [GetDowntime.php](https://github.com/ElliottKasoar/gocdb/blob/sortAPI/lib/Gocdb_Services/PI/GetDowntime.php) as an example, the "bulk" of the sorting - ensuring the downtimes are displayed in oldest first is handled [in the database](https://github.com/ElliottKasoar/gocdb/blob/sortAPI/lib/Gocdb_Services/PI/GetDowntime.php#L226)? This change ensures the endpoints in a single downtime are sorted consistently?

lib/Gocdb_Services/PI/GetDowntime.php
ElliottKasoar: In general, I'd expect `$service['endpointLocations']` to be unsorted, so would need sorting added via `orderArrById()`. However, each `$serviceEndpoint` is only used to populate `$currentServiceEndpointIDs`, and the order of that doesn't matter, as the entire array is considered: `if (in_array($dtEndpoint['id'], $currentServiceEndpointIDs))`

ElliottKasoar:
> @ElliottKasoar: I presume this doesn't enforce the current "ordering"? But rather a sane alternative that will be consistent across DB technologies?

Yes, that's right.

My initial intention was actually to recreate the current "ordering", which I could impose locally to compare to the production, but I don't think that is possible. Even for a specific set of entities in a specific method, there does not appear to be a consistent order between parent entities (e.g. different users' roles are "ordered" differently by get_user, and this "order" is also changed by adding filters).

As a long term change rather than for my testing, using IDs is probably a more sensible solution even if it was possible to recreate the current order.

> Ah, so using [GetDowntime.php](https://github.com/ElliottKasoar/gocdb/blob/sortAPI/lib/Gocdb_Services/PI/GetDowntime.php) as an example, the "bulk" of the sorting - ensuring the downtimes are displayed in oldest first is handled [in the database](https://github.com/ElliottKasoar/gocdb/blob/sortAPI/lib/Gocdb_Services/PI/GetDowntime.php#L226)? This change ensures the endpoints in a single downtime are sorted consistently in the API output?

Yes, more or less. I had to add "bulk" sorting to a couple of methods where it was missing/inconsistent, but the majority of my changes are for "sub-entities" like endpoints.

ElliottKasoar:
(Expanded comments as discussed with @gregcorbett and squashed)

gregcorbett:
> Before merging this into prod, we should announce it to the community as it changes the behaviour of the API

I have captured this here: https://github.com/GOCDB/gocdb/milestone/13

gregcorbett:
rebasing

PR #299, opened 2021-08-20
Add scripts for API comparison

ElliottKasoar:
These are a collection of scripts that can be useful when comparing the API output from two databases. Overall, the scripts are designed to access each API method from both databases and save the results as XML files, carry out diffs of the XML files, and count the length of the diffs. This functionality is largely carried out by get_xml.sh, get_diff.sh and line_count.sh, respectively.

Ideally, both databases should be accessed via the same host, so `GOCDB_PORTAL_URL` is the same for both, minimising differences. In this case, remove_tags.sh can be ignored. However, if using two different hosts, remove_tags.sh can be used to cut lines containing <GOCDB_PORTAL_URL> from XML files. This is not a particularly elegant solution, but as these tags should not contain any unique information, it should not hurt the validity of comparison significantly.

If using the same host, switch_to_mariadb.sh and switch_to_oracle.sh allow switching between two databases. They assume maria_database_connection.php and database_connection.php store the connection details for MariaDB and Oracle respectively. They also update current_db.sh, which simply echos the current database based on the last switch that was called.

All required scripts can be run via run_scripts.sh. This can be used to define and create all directories that are used by other scripts, and optionally allows remove_tags.sh to be correctly used in combination with the other scripts.

The main change required to use these scripts is to update `$oracle_URL` and `$mariadb_URL` in get_xml.sh to the correct host URLs, as well as potentially the wget options depending on host details. It may also be useful/necessary to update the get_downtime and get_downtime_nested_services method calls defined in get_xml.sh. The full files with no parameters would be too large to access, so currently only downtimes with start dates from 01/01/2021 are requested. Otherwise, run_scripts.sh should hopefully run everything without issues.

Temporary text files are written and removed, so be careful where/how these scripts are run.

Note: A significant number of differences will be flagged if the API is not sorted consistently (see #298).

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,70 @@
+API Comparison Scripts
+===========================
+
+* [Setup](#setup)
+* [Running all scripts](#allscripts)
+* [Running individual scripts](#individualscripts)
+* [Notes](#notes)
+
+## Setup <a id="setup"></a>
+* URLs for two GOCDB hosts must be defined in `get_xml.sh` by `oracle_URL` and `mariadb_URL`
+* A certificate and private key must be placed in `/etc/grid-security/hostcert/`, for accessing protected API methods
+* Lists of API methods, their protection levels, and unique file names in `get_xml.sh` should be updated
+  * Defined by `arr_methods_all`, `arr_permissions_all` and `arr_files_all`, and/or individual flags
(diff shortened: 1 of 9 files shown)


gregcorbett:
I think these scripts were very useful for a time, but perhaps aren't needed now.

PR #171, opened 2019-08-14
Fix email issues

jrha:
Depends on #238 

Fixes #64, fixes #163 and fixes #168.

PR #305, opened 2021-09-20
Remove inactive users

pjcon:
Users who have a lastLoginDate from more than 18 months ago will be removed.

Also catch the case where users who have not logged in since the addition of lastLoginDate, which means they don't have a lastLoginDate. They are also removed.

~Depends on~ #171 is now merged into dev
~Merge after #284~

### DIFF
--- a/resources/RemoveInactiveUsersRunner.php
+++ b/resources/RemoveInactiveUsersRunner.php
@@ -73,7 +73,7 @@ function sendWarningEmail($user){
 
     // Email content
     $headers = "From: GOCDB <gocdb-admins@mailman.egi.eu>";
-    $subject = "GocDB: User account deletion notice";
+    $subject = "GOCDB: User account deletion notice";
 
     //$webPortalURL = "gocdb-portal-address";
     $localInfoLocation = __DIR__ . "/../config/local_info.xml";
--- a/resources/RemoveInactiveUsersRunner.php
+++ b/resources/RemoveInactiveUsersRunner.php
@@ -2,6 +2,7 @@
 require_once dirname(__FILE__) . "/../lib/Doctrine/bootstrap.php";
 require dirname(__FILE__) . '/../lib/Doctrine/bootstrap_doctrine.php';
 require_once dirname(__FILE__) . '/../lib/Gocdb_Services/User.php';
+require_once dirname(__FILE__) . '/../lib/Gocdb_Services/Factory.php';
 
 $em = $entityManager;
 $dql = "SELECT u FROM User u";
@@ -56,5 +57,23 @@ function deleteUser($user){
 }
 
 function sendWarningEmail($user){
-    echo "Email sent.";
+    $emailAddress = $user->getEmail();
+    $certDn = $user->getCertificateDn();
+
+    // Email content
+    $headers = "From: no-reply@goc.egi.eu";
+    $subject = "GocDB: User account deletion notice";
+
+    //$webPortalURL = "gocdb-portal-address";
+    $localInfoLocation = __DIR__ . "/../config/local_info.xml";
+    $localInfoXML = simplexml_load_file ( $localInfoLocation );
+    $webPortalURL = $localInfoXML->local_info->web_portal_url;
+   
+    $body = "Dear GOCDB User,\n\n" . "Your account (ID:". $certDn .") has not"
+            . "signed in for the past 17 months and is due for deletion in 30"
+            . "days.\n\n" . "You can prevent this by visiting" . $webPortalURL
+            . "while authenticated with the above credential.\n\n";
+   
+    // Handle all mail related printing/debugging
+    \Factory::getEmailService()->send($emailAddress, $subject, $body, $headers);
 }
--- a/resources/RemoveInactiveUsersRunner.php
+++ b/resources/RemoveInactiveUsersRunner.php
@@ -72,7 +72,7 @@ function sendWarningEmail($user){
     $certDn = $user->getCertificateDn();
 
     // Email content
-    $headers = "From: no-reply@goc.egi.eu";
+    $headers = "From: GOCDB <gocdb-admins@mailman.egi.eu>";
     $subject = "GocDB: User account deletion notice";
 
     //$webPortalURL = "gocdb-portal-address";
--- a/resources/RemoveInactiveUsersRunner.php
+++ b/resources/RemoveInactiveUsersRunner.php
@@ -26,7 +26,8 @@ foreach ($users as $user) {
         echo "Deleting user.\n";
         deleteUser($user, $entityManager);
         echo "\n";
-        return;
+        // Move onto the next users.
+        continue;
     }
 
     $elapsedMonths = (int) $interval->format('%a') / 30;
--- a/resources/RemoveInactiveUsersRunner.php
+++ b/resources/RemoveInactiveUsersRunner.php
@@ -36,7 +36,7 @@ foreach ($users as $user) {
 
     if ($elapsedMonths > 18){ // Delete user
         echo "Deleting user\n";
-	    deleteUser($user);
+	    deleteUser($user, $em);
     } 
     elseif ($elapsedMonths > 17){ // Warn user
         echo "Requesting user warning email.\n";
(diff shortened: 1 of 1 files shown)


resources/RemoveInactiveUsersRunner.php
```diff
@@ -0,0 +1,91 @@
...
+    } catch (\Exception $e) {
+        $em->getConnection()->rollback();
+        $em->close();
+        echo "User not deleted.\n";
+        throw $e;
+    }
+}
+
+function sendWarningEmail($user){
+    $userEmail = $user->getEmail();
```
gregcorbett, line 72:
```suggestion
    $emailAddress = $user->getEmail();
```

resources/RemoveInactiveUsersRunner.php
```diff
@@ -0,0 +1,91 @@
...
+
+    // Email content
+    $headers = "From: no-reply@goc.egi.eu";
+    $subject = "GocDB: User account deletion notice";
+
+    $localInfoLocation = __DIR__ . "/../config/local_info.xml";
+    $localInfoXML = simplexml_load_file ( $localInfoLocation );
+    $webPortalURL = $localInfoXML->local_info->web_portal_url;
+   
+    $body = "Dear GOCDB User,\n\n" . "Your account has not signed in for the past 17 months and is due for deletion in 30 days.\n\n" .
```
gregcorbett, line 82: It would be useful to include the ID String of the account that is set to expire, as a few users will have several (using X.509 DNs or EGI Check-In ID).

As a note to future Greg, when account linking is a thing - this part of the code may need rethinking.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -0,0 +1,92 @@
...
+
+    // Email content
+    $headers = "From: no-reply@goc.egi.eu";
+    $subject = "GocDB: User account deletion notice";
+
+    $localInfoLocation = __DIR__ . "/../config/local_info.xml";
+    $localInfoXML = simplexml_load_file ( $localInfoLocation );
+    $webPortalURL = $localInfoXML->local_info->web_portal_url;
+   
+    $body = "Dear GOCDB User,\n\n" . "Your account (ID:". $accountId .") has not signed in for the past 17 months and is due for deletion in 30 days.\n\n" .
```
gregcorbett, line 83: Should be the users certDn property rather than the ID the user happens to be in the DB.

This line is also quite long and wraps around when comparing diffs side to side. Please split it across mutliple lines.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -0,0 +1,92 @@
...
+        echo "Deleting user\n";
+	    deleteUser($user, $em);
+    } 
+    elseif ($elapsedMonths > 17){ // Warn user
+        echo "Requesting user warning email.\n";
+	    sendWarningEmail($user);
+    } 
+    elseif ($elapsedMonths < 17){ // Do Nothing
+        echo "Doing nothing.\n";
+	    doNothing();
```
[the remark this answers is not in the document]
gregcorbett: i think given the function itself is a one-liner, and is only called once, you could just put the `echo` statement here,

resources/RemoveInactiveUsersRunner.php
gregcorbett: Looks like this has been solved.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -0,0 +1,92 @@
...
+    // Email content
+    $headers = "From: no-reply@goc.egi.eu";
+    $subject = "GocDB: User account deletion notice";
+
+    $localInfoLocation = __DIR__ . "/../config/local_info.xml";
+    $localInfoXML = simplexml_load_file ( $localInfoLocation );
+    $webPortalURL = $localInfoXML->local_info->web_portal_url;
+   
+    $body = "Dear GOCDB User,\n\n" . "Your account (ID:". $accountId .") has not signed in for the past 17 months and is due for deletion in 30 days.\n\n" .
+            "You can prevent this by simply signing in at:\n\n" . $webPortalURL . "\n\n";
```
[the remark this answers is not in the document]
gregcorbett: Looks like this has been solved.

resources/RemoveInactiveUsersRunner.php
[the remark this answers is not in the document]
gregcorbett: Looks like this has been solved.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -0,0 +1,92 @@
...
+
+    $lastLoginDate = $user->getLastLoginDate();
+
+    if ($lastLoginDate){ // null lastLoginDate check
+        $interval = $today->diff($lastLoginDate);
+    } else { // This might only be run once, since new users always have field filled.
+        echo "User has no lastLoginDate (it may have been a very long time.)\n";
+        echo "Deleting user.\n";
+	    deleteUser($user, $em);
+        echo "\n";
```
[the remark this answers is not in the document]
gregcorbett: Fixed in https://github.com/GOCDB/gocdb/pull/305/commits/d0a092f9790d2cc069a82494d333ef6fc7c5493e

resources/RemoveInactiveUsersRunner.php
gregcorbett: This is now done.

gregcorbett:
I've probably touched this too much to review it, can someone else from @GOCDB/gocdb-devs review it please?

gregcorbett:
Can someone from @GOCDB/gocdb-devs review this please?

GRyall:
That's very chatty (echo's content for every user), but maybe that is fine given the subject matter. 

I assume there is a transition process before it is first run to give everyone >18 months a warning?

resources/RemoveInactiveUsersRunner.php
```diff
@@ -0,0 +1,92 @@
...
+    $emailAddress = $user->getEmail();
+
+    // Email content
+    $headers = "From: GOCDB <gocdb-admins@mailman.egi.eu>";
+    $subject = "GOCDB: User account deletion notice";
+
+    $body = "Dear ". $user->getForename() .",\n\n" .
+            "Your GOCDB account, associated with the following identifiers, " .
+            "has not been signed into during the last 17 months and will be " .
+            "when this period of inactivity reaches 18 months.\n\n";
```
ineilson, line 77:
```suggestion
            "deleted when this period of inactivity reaches 18 months.\n\n";
```

ineilson (commented):
There is now (#284 https://github.com/GOCDB/gocdb/blob/2bccf160c2ec2dd429351b0e28eb29ec3305d6b0/lib/Gocdb_Services/User.php#L441) a condition where a user should not be deleted if they are linked with an API authentication credential. Will the delete fail in this case or cause the deletion of the credential, which may be undesirable?

gregcorbett:
Good spot, it will need testing.

#284 doesn't provided a means of cleaning up old/inactive API authentication credentials does it? Maybe in the short term we refuse to delete a user with an API authentication credential.

ineilson:
> #284 doesn't provided a means of cleaning up old/inactive API authentication credentials does it? Maybe in the short term we refuse to delete a user with an API authentication credential.

No, it forces the reassignment/deletion of the API credential before the user can be removed. 

Short term: that sounds good given the limited API cred usage, but there should be some monitoring/log of failures from the automated script to allow follow-up.

ineilson (approved):

gregcorbett:
Need to rebase this to pick up the changes from #284.

gregcorbett:
> There is now (#284
> 
> https://github.com/GOCDB/gocdb/blob/2bccf160c2ec2dd429351b0e28eb29ec3305d6b0/lib/Gocdb_Services/User.php#L441

@ineilson: I've address the above now - in a97fc223c7aa7dbb18dda27b06279ac9e2a621bc

@tofu-rocketry: Following our discussions about how frequently this should run / how much warning a user should have get, I've softcoded thresholds for warning and deletion - in e86f594c02013afadd35d9ba6f89ae82b6fe155b

gregcorbett:
> Looks okay. If a deletion fails for any user for some reason the whole thing stops. That's intended behaviour?

maybe "intended" is to strong a word, but it feels like an okay thing to do.

PR #321, opened 2022-01-17
Take monitoring error info URL from local_info.xml

### LINKED ISSUE #312: portal/GOCDB_monitor/check.php has a hard coded reference to goc.egi.eu
The page should reference the local_info.xml to determine the correct URL to display.

https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/GOCDB_monitor/check.php#L26

ineilson:
Fixes #312

### DIFF
--- a/htdocs/web_portal/GOCDB_monitor/check.php
+++ b/htdocs/web_portal/GOCDB_monitor/check.php
@@ -6,12 +6,12 @@ require_once "tests.php";
 
 $res[1] = test_db_connection();
 $res[2] = test_url(
-            Factory::getConfigService()->GetPiUrl().
+            \Factory::getConfigService()->GetPiUrl().
             get_testPiMethod()
             );
 //$res[3] = test_url(PORTAL_URL);
 $res[3] = test_url(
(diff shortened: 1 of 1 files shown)


htdocs/web_portal/GOCDB_monitor/check.php
```diff
@@ -1,6 +1,9 @@
 <?php
 require_once "tests.php";
 
+// Initialise local_info.xml configuration overrides
+\Factory::getConfigService()->setLocalInfoOverride($_SERVER['SERVER_NAME']);
```
tofu-rocketry, line 5: The backslash is about namespacing, right? Is it needed here?

ineilson: Yes it's namespace spec and no, it's not needed here I think. However, as it is the most common usage in GOCDB codebase, I'll keep as is but make other references in this module follow the same.

gregcorbett (approved):
LGTM :)

gregcorbett:
Just rebasing this so the test run against it.

PR #324, opened 2022-02-03
Refer to X.509, not (x|X)509

gregcorbett:
It's more technically correct, (the best kind of correct), and brings the naming of the X.509 API credential in line with the X.509 user credential (https://github.com/GOCDB/gocdb/pull/294).

so we should merge/release it:
  - after https://github.com/GOCDB/gocdb/pull/260 as this PR depends on that one (will need a rebase).
  - maybe after https://github.com/GOCDB/gocdb/pull/294
  - after, but the same release, as https://github.com/GOCDB/gocdb/pull/284, as this does "break" API access for existing credentials - that is until they next get edited, which users will have to do as a result of 284 anyway.
  - ~After #368~ or not, as this is ready for review.

### DIFF
--- a/config/gocdb_schema.xml
+++ b/config/gocdb_schema.xml
@@ -604,7 +604,7 @@
         <field>
             <fname>TYPE</fname>
             <length>255</length>
-            <regex>/^(X.509|OIDC Subject)$/</regex>
+            <regex>/^(X\.509|OIDC Subject)$/</regex>
         </field>
 		<field>
 			<fname>ALLOW_WRITE</fname>


gregcorbett:
As @ineilson [said](https://github.com/GOCDB/gocdb/pull/260/files#r795599041) in #260, a follow up pass of `(x|X)509 -> X.509` will be needed after #284.

gregcorbett:
I've rebased this on top of the latest dev, but it still needs a second pass.

gregcorbett:
I have now done a second, wider, pass.

config/gocdb_schema.xml
```diff
@@ -604,7 +604,7 @@
         <field>
             <fname>TYPE</fname>
             <length>255</length>
-            <regex>/^(X509|OIDC Subject)$/</regex>
+            <regex>/^(X.509|OIDC Subject)$/</regex>
```
ineilson, line 607:
```suggestion
            <regex>/^(X\.509|OIDC Subject)$/</regex>
```

htdocs/web_portal/components/Get_User_Principle.php
```diff
@@ -137,7 +137,7 @@ function Get_User_AuthType() {
 }
 
 /**
- * Get the user's principle string (x509 DN from certificate or from SAML attribute).
+ * Get the user's principle string (X.509 DN from certificate or from SAML attribute).
```
ineilson, line 140: I believe what we should be referring to here, and probably throughout the code, is _principal_ not _principle_, but, if correct, it's rather late to change it now.

lib/Gocdb_Services/User.php
```diff
@@ -540,6 +540,8 @@ public function getAuthTypes($reducedRealms=true) {
             if (strpos($authTokenName, 'Shib') !== false) {
```
ineilson, line 540: Not from changes but it seems a bit lazy to go to the trouble of creating unique, name-spaced authentication class identifiers and then just check the short substrings _Shib_, _X509_ and _IAM_.

ineilson (commented):
We should probably be correct (if I'm right) for the regex.

htdocs/web_portal/components/Get_User_Principle.php
gregcorbett: Tbh 🤷... In #368, I opted for `identifier`, because that's how we refer them while speaking.

lib/Gocdb_Services/User.php
gregcorbett: Added to #360

gregcorbett:
I'm resisting tackling the codeclimate issues as part of this PR because it's like pulling a thread from a tapestry.

tofu-rocketry:
The commit message on 3e6f7bee75930696d7601ad0546f08e4889c5b37 doesn't at all match or explain what the commit is actually doing.

config/gocdb_schema.xml
```diff
@@ -604,7 +604,7 @@
         <field>
             <fname>TYPE</fname>
             <length>255</length>
-            <regex>/^(X509|OIDC Subject)$/</regex>
+            <regex>/^(X\.509|OIDC Subject)$/</regex>
```
tofu-rocketry, line 607: What does this impact?

gregcorbett:
> The commit message on [3e6f7be](https://github.com/GOCDB/gocdb/commit/3e6f7bee75930696d7601ad0546f08e4889c5b37) doesn't at all match or explain what the commit is actually doing.

I have changed this commit message, it's now 4cbcdcaf104daa417bda58177643391f5b24ed6b.

config/gocdb_schema.xml
gregcorbett: This (and the related changes further "up" in the code) mean that a user would have to edit any existing API authentication entities they still want to use in order for them to pick up the new "X.509" type.

Because of #284, users will have to edit existing API authentication entities following the release of GOCDB 5.10.0 anyway, to associate the entities with a user.

tofu-rocketry (approved):

gregcorbett:
Rebased to resolve conflicts

PR #330, opened 2022-02-14
GT-95 Include failing test and error string in monitoring message

### LINKED ISSUE #223: ops_monitor_check.php message on error ignores which test is failing
`htdocs/web_portal/GOCDB_monitor/ops_monitor_check.php` will show the error message

> GOCDB Web Portal is unable to connect to the GOCDB back end database\n

regardless of which test is failing (it could be a API problem, a portal problem or a problem around the local_info.xml). It could give a more generic error message like the `htdocs/web_portal/GOCDB_monitor/check.php` page does. One may then ask "why do we have two very similiar pages?"

ineilson:
Fixes #223.

### DIFF
--- a/htdocs/web_portal/GOCDB_monitor/ops_monitor_check.php
+++ b/htdocs/web_portal/GOCDB_monitor/ops_monitor_check.php
@@ -1,49 +1,22 @@
 <?php
 require_once "tests.php";
 
-// function returns associative array
-$res[1] = test_db_connection();
-
-$res[2] = test_url(Factory::getConfigService()->GetPiUrl().get_testPiMethod());
-
-$res[3] = test_url(Factory::getConfigService()->GetPortalURL());
-
-$counts=array(
-        "ok" => 0,
-        "warn" => 0,
-        "error" => 0
-);
-
-foreach ($res as $r){
-    $counts[$r["status"]]++;
-}
-/*
-If there is an error, counts["error"] is incremented, e.g.
-Array
-(
-    [ok] => 0
-    [warn] => 0
-    [error] => 1
-)
- */
-
-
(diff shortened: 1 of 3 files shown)


gregcorbett (changes requested):
LGTM, please rebase this to fix the conflicts 😄

htdocs/web_portal/GOCDB_monitor/ops_monitor_check.php
```diff
@@ -1,49 +1,22 @@
...
-    die();
+if (isset($_REQUEST['fake_failure'])) {
+    $message = 'Fake failure';
+    $errorCount = 1;
+} else {
+    $errorCount = run_tests($message);
 }
 
-if ($counts["error"] != 0) {
+if ($errorCount != 0) {
```
gregcorbett, line 20: How does `$errorCount` get populated? In `check.php`, there is this line

```
$errorCount = run_tests($message);
```

 `ops_monitor_check.php` doesn't seem to call that method.

gregcorbett: ah you're right, didn't see that bit. Which perhaps suggests some additional comment is needed.

Maybe something like

```
/* If someone wants to test the failure, they can fake one using
 * the fake_failure parameter. 
 * Otherwise, run the tests. */
```

ineilson: There is already 'something like' that in the header which seems to me a better place.

gregcorbett (approved):
LGTM

gregcorbett:
rebasing on top of latest dev

PR #334, opened 2022-02-15
Update composer.lock

gregcorbett:
via `composer update`. Nothing obvious immediately broke.

Before merging, this should probably be updated with another `composer update` to see if anything has changed in the meantime.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,9 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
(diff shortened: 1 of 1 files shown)


GRyall:
Are we at the maximum versions of things for the version of php we are running? Is it worth looking at whether we can bump up some of the versions of the software we are running e.g. doctrine version? Given that the ORM version we are using is now unmaintained - https://www.doctrine-project.org/projects/orm.html

gregcorbett:
Closing in favour of #343 .

PR #346, opened 2022-05-17
Release 5.9.0 to dev

gregcorbett:
It's running here: https://host-172-16-101-131.nubes.stfc.ac.uk/portal

PR #348, opened 2022-05-19
Add support for EOSC IdPs out of the box

gregcorbett:
Use "voPersonID" to consume the user's identifier, rather than ePUID - as that has been deprecated following the (AAI teams?) adoption of the AARC-G026 guidelines.

Depends on #346


gregcorbett:
Note to Future Greg: some of the 62 codeclimate issues are low hanging fruit and should be fixed before marking as ready for review.

gregcorbett:
This entitlement used in this will also need to be changed to the new one provided by EOSC AAI.

gregcorbett:
> This entitlement used in this will also need to be changed to the new one provided by EOSC AAI.

This is now done.

gregcorbett:
Superseded by #430

PR #290, opened 2021-07-15
Configurable AUP and privacy notice

ineilson:
Closes #285

PR #353, opened 2022-06-01
Update sponsor logos and text

gregcorbett:
EDIT: no longer WIP

This is still a WIP. That said, people can eyeball it and suggest improvements if they want. It's running here: https://host-172-16-102-204.nubes.stfc.ac.uk

Still to do:
- [x] use an "official" EOSC Future logo, rather than one I screenshotted from the brand guidelines
- [x] use the transparent IRIS Logo throughout
- [x] I want to merge #290 first to spare @ineilson another rebase

Work on movnig these to a footer will be seperate PR, being put together here: https://github.com/gregcorbett/gocdb/tree/logo_footer


gregcorbett:
> Looks good but perhaps crowded.

Yeah, that's what made me start to think about putting them in a footer, because that side bar is getting long...

> Can we drop the EU logo now as we're not directly funded (I think).

Ah yes, I think before EOSC-hub part of the project funding did come "direct" from the EC. That's not the case now, atleast not in a way we see.

> There is an EGI-ACE logo - https://www.egi.eu/project/egi-ace/ - use that instead.

Yeah... thats probably better...

> Should say STFC - UKRI, not just STFC (or is it the other way around?).

👍 

> Perhaps we don't need the text list of sponsors, just logos would be enough? (with alt-text)

I quite like the text, i think it provides a bit of clarity as to why the logos have been included.

gregcorbett:
> > There is an EGI-ACE logo - https://www.egi.eu/project/egi-ace/ - use that instead.
> 
> Yeah... thats probably better...

MattV will look into this and confirm which logo to use.

gregcorbett:
> > There is an EGI-ACE logo - https://www.egi.eu/project/egi-ace/ - use that instead.
> 
> 
> Yeah... thats probably better...

Nay, we should be using the generic one.

> > Can we drop the EU logo now as we're not directly funded (I think).
> 
> 
> Ah yes, I think before EOSC-hub part of the project funding did come "direct" from the EC. That's not the case now, atleast not in a way we see.

Nay, GOCDB benefits from other EU projects - such as EOSC Future, so we need to leave it in.

gregcorbett:
> Should say STFC - UKRI, not just STFC (or is it the other way around?).

So the first time we mention STFC, we should use "STFC, part of UK Research and Innovation" - as we do on the landing page.

Subsequent mentions can revert to just the council name. Maybe when the logos are in a footer we can match the usage on the landing page.

gregcorbett:
Rebasing on latest dev to resolve conflicts.

gregcorbett:
Several rebases later, this is ready for review @GOCDB/gocdb-devs.

tofu-rocketry (approved):

PR #354, opened 2022-06-09
Fix JS regex use to validate IRIS IAM strings

gregcorbett:


htdocs/web_portal/javascript/linking.js
```diff
@@ -85,7 +85,7 @@ function getRegExIdString() {
     // End with @iris.iam.ac.uk
     } else if (authType === "IRIS IAM - OIDC") {
         // var regExIdString = /^([a-f0-9]{8}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{12})@iris\-iam\.ac\.uk$/;
-        var regExIdString = /^.+@iris\-iam\.ac\.uk$/;
+        var regExIdString = /^.+@iris\-iam\.stfc\.ac\.uk\/$/
```
ineilson, line 88: My regex is rusty but did you intend to force a trailing slash? Added line termination semi-colon.
```suggestion
        var regExIdString = /^.+@iris\-iam\.stfc\.ac\.uk$/;
```

gregcorbett: I do mean to force a trailing slash, it's included in the ID string the IRIS IAM gives out. Good spot about the `;` :)

ineilson: For a future change maybe but this very much feels like it shouldn't be hardcoded.

ElliottKasoar:
Minor point but should `// End with @iris.iam.ac.uk` be updated to reflect the correct string?

It's probably obvious, but just to note that the commented out `regExIdString` above was planned to check the rest of the form of the ID string. It may be worth either implementing it for a more detailed check, or removing it if it's never going to be used (including it would "give away" the full form of the ID string, which may be a issue in itself).

ineilson:
> Minor point but should `// End with @iris.iam.ac.uk` be updated to reflect the correct string?

Seems like a good idea.

ineilson:
> It's probably obvious, but just to note that the commented out `regExIdString` above was planned to check the rest of the form of the ID string. It may be worth either implementing it for a more detailed check, or removing it if it's never going to be used (including it would "give away" the full form of the ID string, which may be a issue in itself).

Why was it not used in the first place, and what would be the implication of the 'give away'?

ElliottKasoar:
> Why was it not used in the first place

From what I can recall, I think it was a combination of seeming unnecessary, and that I couldn't test it, as my instance of GOCDB didn't have IRIS IAM auth enabled (so I wasn't even registered myself).

I think the longer form was essentially a placeholder based on discussions with @gregcorbett about the expected full form of the ID string, but later we moved towards more minimal validation that was sufficient to differentiate the different auth types (e.g. `// End with @egi.eu` and `// Start with slash only`).


> and what would be the implication of the 'give away'?

I meant in the sense of helping people too much in guessing what an ID string should be that they're not in possession of, but I don't think it's really an issue?

gregcorbett:
> Minor point but should // End with @iris.iam.ac.uk be updated to reflect the correct string?

Ah yes, I've changed the code but not the comment 🤦

gregcorbett:
> > Minor point but should // End with @iris.iam.ac.uk be updated to reflect the correct string?
> 
> Ah yes, I've changed the code but not the comment 🤦

I've fixed this and rebased (into a single commit) the branch on top of the latest dev.

gregcorbett:
This has been running on the IRIS instance since at least Jun 2022.

htdocs/web_portal/javascript/linking.js
```diff
@@ -84,8 +84,8 @@ function getRegExIdString() {
 
     // End with @iris.iam.ac.uk
     } else if (authType === "IRIS IAM - OIDC") {
-        // var regExIdString = /^([a-f0-9]{8}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{12})@iris\-iam\.ac\.uk$/;
-        var regExIdString = /^.+@iris\-iam\.ac\.uk$/;
+        // var regExIdString = /^([a-f0-9]{8}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{4}\-[a-f0-9]{12})@iris\-iam\.stfc\.ac\.uk$/;
+        var regExIdString = /^.+@iris\-iam\.stfc\.ac\.uk\/$/;
```
Sae126V, line 88: Just a clarification, Greg. So this now would be having / (relative URL) in the end?

gregcorbett: Yep 😄

Sae126V (approved):
LGTM

htdocs/web_portal/javascript/linking.js
gregcorbett: it's a weridness of the IRIS IAM.

gregcorbett:
rebasing on latest dev

Issue #270, opened 2021-02-10
Explicitly check for code style compliance against some standard at PR time


gregcorbett:
It's not clear to me what codacy and codeclimate are actually using as "the gold standard" of PHP code. Specifying a well known code standard is [at least possible in codeclimate](https://docs.codeclimate.com/docs/phpcodesniffer). 

We also have a PHP Mess Detector config ruleset around (https://github.com/GOCDB/gocdb/pull/265), and I like the fact we can use that to configure exceptions for things we are never going to change (variable names like `$em` for example). But with phpcodesniffer it seems we can aim for a standard, even if it tells us "the code is bad and you should feel bad" all the time. Perhaps another PHP linter / CI platform can do both.

[PSR-12: Extended Coding Style ](https://www.php-fig.org/psr/psr-12/) is a standard. Other standards may be available.

tofu-rocketry:
#265 has now been merged to give a basic PHP Mess Detector rule set.

PSR-12 seems like a good standard to aim for as a baseline, as it's used by a number of large PHP projects. I suggest we add a simple `phpcs.xml` for PHP CodeSniffer to check against `PSR12` (as the CodeSniffer option is written). Then we have the following to follow up on:

- **Codacy** will pick up `phpcs.xml` and `phpmd.xml` (from #265) config once you've selected them in the web interface.
- **Codeclimate** needs a config file to pick up a `phpmd.xml` and only seems to have the option to configure codeSniffer from the Codeclimate config.
- **GitHub Actions** could be used to run either/both checkers, but then it would be good to merge #288 first.

gregcorbett:
So after #365, codeSniffer on Codeclimate will check the code against PSR12?

tofu-rocketry:
Ok, Codeclimate knows that the config is coming from a `.codeclimate.yml` so I guess this is now done.

![image](https://user-images.githubusercontent.com/4836233/193555562-4e41ec43-208b-43ab-b6cb-bdfec647fa2a.png)

PR #363, opened 2022-07-04
Add basic PHP CodeSniffer rule set

tofu-rocketry:
Only partially addresses #270, but it will be good to add this basic `phpcs.xml` and see how it goes with Codacy picking it up, then we can decide it we want to add a config for Codeclimate (and/or add Actions).


gregcorbett (approved):
LGTM

PR #368, opened 2022-07-08
Refactor API Auth to support non X.509 identifiers

gregcorbett:
We want the API to allow access based on OIDC Identifiers in the future., so need to generalise getAPIAuthentication.

~Still a work in progress, I should rename the `authByCert` method at least and fix the tests.~

I also wonder wether we should remove [this check](https://github.com/GOCDB/gocdb/blob/5c302472a6a7459dd4f6cd275a4f3c41f6113726/htdocs/PI/index.php#L386), maybe API access should be a function of the API Identifiers alone - not something a user can browse by virtue of their roles?

### DIFF
--- a/lib/Gocdb_Services/APIAuthenticationService.php
+++ b/lib/Gocdb_Services/APIAuthenticationService.php
@@ -37,21 +37,19 @@ class APIAuthenticationService extends AbstractEntityService{
      * Returns the APIAuthentication entity associated with the given identifier.
      *
      * @param string $ident Identifier (e.g. X.509 DN as string)
-     * @param string $type  Identifyer type (e.g. "X.509")
      * @return \APIAuthentication APIAuthentication associated with this identifier
      */
-    public function getAPIAuthentication($ident, $type) {
+    public function getAPIAuthentication($ident) {
 
         if (!is_string($ident)) {
             throw new \LogicException("Expected string APIAuthentication identifier.");
         }
 
         $dql = "SELECT a FROM APIAuthentication a " .
-                "WHERE (a.identifier = :ident AND a.type = :type)" ;
+                "WHERE (a.identifier = :ident)" ;
 
         $qry = $this->em->createQuery($dql);
         $qry->setParameter('ident', $ident);
-        $qry->setParameter('type', $type);
 
         $apiAuth = $qry->getOneOrNullResult();
 
--- a/tests/unit/lib/Gocdb_Services/APIAuthenticationServiceTest.php
+++ b/tests/unit/lib/Gocdb_Services/APIAuthenticationServiceTest.php
@@ -172,29 +172,31 @@ class APIAuthEnticationServiceTest extends PHPUnit_Extensions_Database_TestCase
         $type = 'X.509';
       // Start with no APIAuthentication entities to be found
         $this->assertNull(
-            $authEntServ->getAPIAuthentication($ident, $type),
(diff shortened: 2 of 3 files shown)


gregcorbett:
> I also wonder wether we should remove [this check](https://github.com/GOCDB/gocdb/blob/5c302472a6a7459dd4f6cd275a4f3c41f6113726/htdocs/PI/index.php#L386), maybe API access should be a function of the API Identifiers alone - not something a user can browse by virtue of their roles?

Admins should probably be able to access the API by virtue of being admins. 

The [message](https://github.com/GOCDB/gocdb/blob/4b9fccfe4565b54fd5aa493c362f173155baee4b/htdocs/PI/index.php#L413) when an unauthorized request comes in should be changed. If it's a robot request it's not apporpriate to return "get a role", it should be along the lines of "a human should add your robot API identifier". This will probably be simplier if we remove the check the user ID check.

gregcorbett:
> I also wonder wether we should remove [this check](https://github.com/GOCDB/gocdb/blob/5c302472a6a7459dd4f6cd275a4f3c41f6113726/htdocs/PI/index.php#L386), maybe API access should be a function of the API Identifiers alone - not something a user can browse by virtue of their roles?

Hmm... maybe not... we were vague and said we would "restrict the access to personal data to users with roles at sites or registered robots", no distinctions for access personal data in the portal or the API...

gregcorbett:
I've rebased this over the conflict, to avoid a 3rd pass of #324, I will merge that one first.

gregcorbett:
Converted to draft until #324 is merged.

gregcorbett:
Rebased to fix the conflicts.

lib/Gocdb_Services/APIAuthenticationService.php
```diff
@@ -37,21 +37,19 @@ function __construct() {
      * Returns the APIAuthentication entity associated with the given identifier.
      *
      * @param string $ident Identifier (e.g. X.509 DN as string)
-     * @param string $type  Identifyer type (e.g. "X.509")
      * @return \APIAuthentication APIAuthentication associated with this identifier
      */
-    public function getAPIAuthentication($ident, $type) {
```
tofu-rocketry, line 43: How come you're removing the type?

gregcorbett: So that the code doesn't have to determine the type of the incoming identifier. It's sufficent just to look for the identity string.

tests/unit/lib/Gocdb_Services/APIAuthenticationServiceTest.php
```diff
@@ -172,29 +172,31 @@ public function testGetAPIAuthentication()
...
 
         $authEnt = $siteService->addAPIAuthEntity(
             $site,
             $user,
-            array('IDENTIFIER' =>  $ident,
-                              'TYPE' => $type,
-            'ALLOW_WRITE' => false)
+            array(
+                'IDENTIFIER' =>  $ident,
+                'TYPE' => $type,
```
tofu-rocketry, line 185: Where will `$type` come from?

gregcorbett: Line 172, it's not been removed when adding an API Identifier so that the approporiate validation can be run against it.

tofu-rocketry (approved):

PR #375, opened 2022-08-02
Prevent database connection pile-up during tests

ineilson:
Add tearDown function to tests to properly close the
connection to the database with failure due to >150
connections default max value for MariaDB.

Auto-reformat code.


gregcorbett (approved):
LGTM

PR #380, opened 2022-08-23
Re-add authByAnyCert/Identifier method

gregcorbett:
~~Depends on #368~~

So some methods can be accessed with just a valid identifier. 

i.e. `get_cert_status_date`, for GOCDB 5.9 it's accessible with just a valid identifier and it doesn't contain personal/contact so it should remain accessible with just a valid identifier.

### DIFF
--- a/htdocs/PI/index.php
+++ b/htdocs/PI/index.php
@@ -362,8 +362,7 @@ class PIRequest {
                     break;
             }
         } catch (\Exception $e) {
-            print_r($e->getMessage());
-            die("An error has occured, please contact the GOCDB administrators at gocdb-admins@egi.eu");
+            die($e->getMessage());
         }
         return $xml;
     }
@@ -416,8 +415,10 @@ class PIRequest {
     function authByAnyIdentifier()
     {
         if (empty($this->identifier)) {
-            die("No valid identifier found. Try accessing the " .
-                "resource through the private interface.");
+            throw new \Exception(
+                "No valid identifier found. Try accessing the resource " .
+                "through the private interface."
+            );
         }
     }
 }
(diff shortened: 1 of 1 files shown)


gregcorbett:
Subject to #368 being reviewed an merged, only 77b9b2a needs reviewing in this PR.

htdocs/PI/index.php
```diff
@@ -414,6 +413,13 @@ function authByCert() {
         }
     }
 
+    function authByAnyIdentifier()
+    {
+        if (empty($this->identifier)) {
+            die("No valid identifier found. Try accessing the " .
```
ineilson, line 419: Why do we die rather than throw an exception as in authByIdentifier()?

gregcorbett: not sure, it's "always" been like that. Given `getXml()` is one giant `try... catch` throwing an exception here seems better.

I think all the other methods in `PI/index.php` now throw an exception, so I might also tweak the logic in the [catch](https://github.com/GOCDB/gocdb/blob/dev/htdocs/PI/index.php#L364) slightly

htdocs/PI/index.php
```diff
@@ -328,7 +328,7 @@ function getXml() {
                     break;
                 case "get_cert_status_date" :
                     require_once($directory . 'GetCertStatusDate.php');
-                    $this->authByCert();
+                    $this->authByAnyIdentifier();
```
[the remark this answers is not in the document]
gregcorbett: See 0e9cc671e53bda27099ca45f994b4ffa03ea36e4 😄

htdocs/PI/index.php
gregcorbett: See 765e9f0601cb554b0495538ddff44c49c18c59a0

ineilson (approved):

ineilson (approved):

PR #383, opened 2022-10-03
Add unordered lists tags to fix wonky list

### LINKED ISSUE #371: Wonky text on account linking/recovery page
Text seems to be trying to escape into the left margin. Happens for both linking and recovery versions.

image

https://goc.egi.eu/portal/index.php?Page_Type=Link_Identity

tofu-rocketry:
Resolves #371.

This nested list was missing list tags on the outer list which was making it appear wonky. Inner scope indented.


gregcorbett (approved):
LGTM 😄

PR #385, opened 2022-10-10
381 user removal dry run

### LINKED ISSUE #381: Add dry run option to inactive user removal script
Would be useful to be able to fetch the list of users to be removed for inspection before committing to the removal.

https://github.com/pjcon/gocdb/blob/a97fc223c7aa7dbb18dda27b06279ac9e2a621bc/resources/RemoveInactiveUsersRunner.php

rowan04:
Resolves #381 

The output:
```[root@host-172-16-113-110 resources]# php RemoveInactiveUsersRunner.php --warning_threshold 4 --deletion_threshold 8 --dry_run
Dry run is on.

Scanning user login dates in database at: Mon, 10 Oct 2022 16:08:51


User ID: 1
Months elapsed since last login: 12.78
This user will be deleted when it isn't a dry run.


User ID: 2
Months elapsed since last login: 1.81
Doing nothing.

User ID: 4
Months elapsed since last login: 5.98
Sending user warning email.

<!--
Sending mail disabled, but would have sent:
From: GOCDB <gocdb-admins@mailman.egi.eu>
To: sample@sample.com
Subject: GOCDB: User account deletion notice

Dear Donna,

Your GOCDB account, associated with the following identifiers, has not been signed into during the last 5 months and will be deleted when this period of inactivity reaches 8 months.

Identifiers:
  - X.509: /C=00/O=00000/OU=000000/CN=Donna Moss

You can prevent the deletion of this account by visiting the GOCDB portal while authenticated with one of the above identifiers.


Additional Parameters:
-->


User ID: 5
Months elapsed since last login: 10.35
This user will be deleted when it isn't a dry run.


User ID: 6
Months elapsed since last login: 0.1
Doing nothing.

View the contents of InactiveUsersToBeDeleted.txt to see the list of users who would have been deleted had this not have been a dry run.

Completed ok: Mon, 10 Oct 2022 16:08:51

[root@host-172-16-113-110 resources]# cat InactiveUsersToBeDeleted.txt
These are the users that would have been deleted during the dry run on: Mon, 10 Oct 2022 16:08:51
The user with Id 1 would have been deleted as they last logged in 12 months ago.
The user with Id 5 would have been deleted as they last logged in 10 months ago.
```

### DIFF
--- a/resources/RemoveInactiveUsersRunner.php
+++ b/resources/RemoveInactiveUsersRunner.php
@@ -35,27 +35,29 @@ if (isset($options["deletion_threshold"])) {
 };
 
 if (isset($options["dry_run"])) {
-    $dryRun = true; // If the dry_run option is added, variable dryRun set to true.
-    echo "Dry run is on.\n";
+    // If the dry_run option is added, variable dryRun set to true.
+    $dryRun = true;
+    echo "Dry run is on.\n\n";
 } else {
-    $dryRun = false; // If the dry_run option isn't added, variable dryRun set to false.
-    echo "Dry run is off.\n";
+    // If the dry_run option isn't added, variable dryRun set to false.
+    $dryRun = false;
+    echo "Dry run is off.\n\n";
 };
 
 $dql = "SELECT u FROM User u";
 $users = $entityManager->createQuery($dql)->getResult();
 
-echo "\nScanning user login dates in database at: " . date('D, d M Y H:i:s') . "\n\n";
+echo "Scanning user login dates in database at: " . date('D, d M Y H:i:s') . "\n\n";
 
-if ($dryRun == true) {  // If dry run option has been selected
-    // Text is overwritten to InactiveUsersToBeDeleted.txt
-    file_put_contents("InactiveUsersToBeDeleted.txt", "userID, lastLoginDate, elapsedMonths\n");
+if ($dryRun == true) {
+    // If dry run option has been selected, text is overwritten to InactiveUsersToBeDeleted.csv
+    file_put_contents("InactiveUsersToBeDeleted.csv", "userID, lastLoginDate, elapsedMonths\n");
 };
 
 $today = new DateTime();
 
 foreach ($users as $user) {
-    echo "\nUser ID: " . $user->getId() . "\n";
+    echo "User ID: " . $user->getId() . "\n";
 
     $creationDate = $user->getCreationDate();
     $creationStr = $creationDate->format('Y-m-d H:i:s');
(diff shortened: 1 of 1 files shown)


resources/RemoveInactiveUsersRunner.php
```diff
@@ -57,36 +72,58 @@
...
         $interval = $today->diff($lastLoginDate);
     } else { // This might only be run once, since new users always have field filled.
-        echo "User has no lastLoginDate (it may have been a very long time.)\n";
-        echo "Deleting user.\n";
+        echo "Deleting this user as it has no last login date " .
+        "(it may have been a very long time). \n";
+        //writing to file to say who would have been deleted
+        file_put_contents("InactiveUsersToBeDeleted.txt", "The user " .
+        "with Id ".$user->getId()." would have been deleted as it " .
+        "has no last login date.\n", FILE_APPEND);
```
gregcorbett, line 80: Similar to the other delete later on, we only want to log to a file in dry run mode - so an if statement is needed here.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -57,36 +72,58 @@
...
 
-    $elapsedMonths = (int) $interval->format('%a') / 30;
-    echo 'Months elapsed since last login: ' . $elapsedMonths . "\n";
-
-    if ($elapsedMonths > $deletionThreshold) { // Delete user
-        echo "Deleting user\n";
-        deleteUser($user, $entityManager);
+    $elapsedMonths = (int) $interval->format('%a') / 30.4375; // 30.4375 is 365.25/12
+    echo 'Months elapsed since last login: ' . round($elapsedMonths,2) . "\n";
+    // Rounded months since last login to 2 decimal places for easier reading.
```
gregcorbett, line 89: this comment should preceed the code if refers to.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -57,36 +72,58 @@
...
+
+    if ($elapsedMonths > $deletionThreshold) // Delete user
+        // Check if it is a dry run
+        if ($dryRun == '1') {
+            // Dry run option is set, so append user to be deleted to the file
+            echo "This user will be deleted when it isn't a dry run. \n\n";
+            file_put_contents("InactiveUsersToBeDeleted.txt", "The user with " .
```
gregcorbett, line 96: For the file output, it would be good if it was more machine readbale - so we can pipe it into other tools or open it in excel.

Maybe something like CSV

```
User, lastLoginDate, elapsedMonths
1, '10/11/2022 17:16:00', 11.00
2, '10/10/2022 17:16:00', 0.00
...
```

It would make the output easier to sort and easier to confirm that it's behaving sensibly.

gregcorbett: same for case where `lastLoginDate` is `null`

```
3, `null`, '?'
```

not really sure what's best to log for `elapsedMonths` when `lastLoginDate` is `null` - '?' doesn't feel to bad...

gregcorbett:
Generally looks good :), I've left a few comments about specific things.

Can you rebase the two commits so that the first commit doesn't comment out code only for the second commit to add it back?

Can you address the issues Codacy has picked up?

resources/RemoveInactiveUsersRunner.php
```diff
@@ -32,21 +34,34 @@
     return;
 };
 
+if (isset($options["dry_run"])) {
+    $dryRun = 1; // If the dry_run option is added, variable dryRun set to 1.
```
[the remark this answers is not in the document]
rowan04: Agreed, makes more sense. Not sure why I didn't use Boolean.

resources/RemoveInactiveUsersRunner.php
rowan04: According to an earlier comment left before I edited, there shouldn't be any users without a last login date, as all new users will have one, and any users without a last login date would have been deleted the first time the program was run (which has already happened? not sure).  But ok will fix.

resources/RemoveInactiveUsersRunner.php
rowan04: ok.

resources/RemoveInactiveUsersRunner.php
rowan04: ok

resources/RemoveInactiveUsersRunner.php
```diff
@@ -57,36 +71,60 @@
     if ($lastLoginDate) { // null lastLoginDate check
         $interval = $today->diff($lastLoginDate);
     } else { // This might only be run once, since new users always have field filled.
-        echo "User has no lastLoginDate (it may have been a very long time.)\n";
-        echo "Deleting user.\n";
+        echo "Deleting this user as it has no last login date " .
+        "(it may have been a very long time). \n";
+        if ($dryRun == true) {  // If dry run option has been selected
```
tofu-rocketry, line 76:
```suggestion
        if ($dryRun == true) {  // If dry run option has been selected
```

tofu-rocketry: Some errant whitespace. You can accept my suggestion to fix it without needed to edit it on your development host.

tofu-rocketry (changes requested):

tofu-rocketry (changes requested):

resources/RemoveInactiveUsersRunner.php
rowan04: ok

resources/RemoveInactiveUsersRunner.php
```diff
@@ -144,3 +183,4 @@ function sendWarningEmail($user, $elapsedMonths, $deletionThreshold)
     // Handle all mail related printing/debugging
     \Factory::getEmailService()->send($emailAddress, $subject, $body, $headers);
 }
+
```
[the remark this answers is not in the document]
rowan04: ok

tofu-rocketry (commented):
Looks ok to me, but I'll let Greg have approve it.

tofu-rocketry (dismissed):

resources/RemoveInactiveUsersRunner.php
```diff
@@ -57,36 +71,60 @@
...
-        echo "User has no lastLoginDate (it may have been a very long time.)\n";
-        echo "Deleting user.\n";
+        echo "Deleting this user as it has no last login date " .
+        "(it may have been a very long time). \n";
+        if ($dryRun == true) {  // If dry run option has been selected
+             //writing to file to say who would have been deleted
+            file_put_contents("InactiveUsersToBeDeleted.txt", $user->getId() . ", `null`, " .
+            "`?`\n", FILE_APPEND);
+        };  // no else needed.
         deleteUser($user, $entityManager);
```
gregcorbett, line 81: `deleteUser` is called here regardless of whether `$dryRun == true` is `true` or not. The method call should be in an `else` clause.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -57,36 +71,60 @@
...
+            " " . $lastLoginDate->format("D d M Y H:i:s") . "," .
+            " " . floor($elapsedMonths) . "\n", FILE_APPEND);
+        } else {  // Delete user as dry run option not set.
+            echo "Deleting user. \n";
+            deleteUser($user, $entityManager);
+        };
     } elseif ($elapsedMonths > $warningThreshold) { // Warn user
-        echo "Sending user warning email.\n";
+        echo "Sending user warning email.\n\n";
         sendWarningEmail($user, $elapsedMonths, $deletionThreshold);
```
gregcorbett, line 105: We don't want to send warning emails in dry run mode.

resources/RemoveInactiveUsersRunner.php
```diff
@@ -57,36 +71,60 @@
...
+    // Rounded months since last login to 2 decimal places on line below for easier reading.
+    echo 'Months elapsed since last login: ' . round($elapsedMonths, 2) . "\n";
 
     if ($elapsedMonths > $deletionThreshold) { // Delete user
-        echo "Deleting user\n";
-        deleteUser($user, $entityManager);
+        // Check if it is a dry run
+        if ($dryRun == true) {
+            // Dry run option is set, so append user to be deleted to the file
+            echo "This user will be deleted when it isn't a dry run. \n\n";
```
gregcorbett, line 95:
```suggestion
            echo "This user will be deleted when it isn't a dry run.\n";
```

One new line is probaby fine here, to match the output on a "Do Nothing".

gregcorbett:
Looks good, a few final general things:

- I'd rather strings to be echo'd didn't start with a `\n`. i.e. "echo "\nScanning user login dates in database at: " . date('D, d M Y H:i:s') . "\n\n";" Can you rework the echo statements so the new lines come at the end of the previous echo statement?
- Can the inline comments you've added, i.e. "if ($dryRun == true) { // Check if it is a dry run" be reworked so the comment is not on the same line as a statement?

resources/RemoveInactiveUsersRunner.php
```diff
@@ -35,88 +35,108 @@
...
-        echo "\n";
-        // Move onto the next users.
+            echo "\n";
+        } else {
+            // As dry run option is false, user can be deleted.
+            deleteUser($user, $entityManager);
+            // Move onto the next users.
+            continue;
+        }
         continue;
```
gregcorbett, line 91: I think we only need 1 `continue` here, along the lines of

```
if {
    ...
} else {
    if (dryrun) { 
        don't delete
     } else { 
        delete
     }
    continue
}
```

gregcorbett (changes requested):
looks good, just one additional change needed.

gregcorbett (approved):
LGTM

PR #393, opened 2022-11-08
373 deploying the sample data fails

rowan04:
Resolves #373

PR #394, opened 2022-11-08
Refreshing sample data

### LINKED ISSUE #384: Refresh sample data so that it is not based on real data
The [sample data](https://github.com/GOCDB/gocdb/tree/dev/lib/Doctrine/deploy/sampleData) looks like it was based on some real data at one point. It would be good to refresh this data so that this is no longer the case.

It would also be an oppurtunity to tweak the structure of the sample data, so that it better mirrors present data - i.e. contains multiple NGIs (one of which isn't under a project) like below (the sites should also have dummy services.):

```
        P1
      /    \    
     N1    N2      N3
    /  \   |       |
   S1  S2  S3      S4
```

Users should be given unique, but still dummy, emails so they are easier to differentiate (i.e. in dry runs of scripts that generate emails).

There exists some ["simple sample data"](https://github.com/GOCDB/gocdb/tree/dev/lib/Doctrine/deploy/simpleSampleData) which would be a good starting point to expand and replace the existing sample data.

As a follow on, the example API output used in the [docs](https://github.com/GOCDB/GOCDB.github.io/tree/main/api/read) should be updated to reflect the new sample data, see: https://github.com/GOCDB/GOCDB.github.io/issues/12.

rowan04:
Resolves #384
- Sanitised data so it's not based on real data
- Added multiple NGIs with sites and services
- Added more users with unique dummy emails


lib/Doctrine/deploy/AddNGIs.php
```diff
@@ -51,6 +56,10 @@
             $securityEmail = (string) $value;
         }
 
+        if((string) $key == "OBJECT_ID") {
```
rowan04, line 59: This is the line causing the Codacy fail - however it is in the same format as the other if statements above. Should I fix?

gregcorbett: As it's a new line of code, could you address the issue. A seperate PR that fixes the issue for the file as a whole would also be most welcome :)

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
```diff
@@ -53,15 +53,15 @@
...
+	<SITENAME>Torch</SITENAME>
         <COUNTRY_NAME>Switzerland</COUNTRY_NAME>
         <COUNTRY_CODE>CH</COUNTRY_CODE>
         <ROC_NAME>NGI_CH</ROC_NAME>
         <URL></URL>
     </SERVICE_ENDPOINT>
     <SERVICE_ENDPOINT PRIMARY_KEY="43873G0">
         <PRIMARY_KEY>43873G0</PRIMARY_KEY>
-        <HOSTNAME>bdii.lhep.unibe.ch</HOSTNAME>
+        <HOSTNAME>bdii.lhep.unilu.grid.ch</HOSTNAME>
```
gregcorbett, line 64: Hmmmm `unilu.ch` is a thing, maybe we could change `unilu.grid.ch` to something else - perhaps `torch.grid.ch`?

gregcorbett: or perhaps something different so we can change the `unilu` sitename to something different to `torch`.

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
```diff
@@ -285,11 +285,385 @@
...
+        <NODE_MONITORED>Y</NODE_MONITORED>
+        <SITENAME>Izola MRF</SITENAME>
+        <COUNTRY_NAME>Slovenia</COUNTRY_NAME>
+        <COUNTRY_CODE>SI</COUNTRY_CODE>
+        <ROC_NAME>NGI_SI</ROC_NAME>
+        <URL></URL>
+    </SERVICE_ENDPOINT>
+    <SERVICE_ENDPOINT PRIMARY_KEY="10007G0">
+        <PRIMARY_KEY>10007G0</PRIMARY_KEY>
+        <HOSTNAME>uni-bdii.uni-lj.si</HOSTNAME>
```
gregcorbett, line 381: looks like `uni-lj.si` is a thing, could this be changed?

gregcorbett: and changed in other places as well.

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
```diff
@@ -285,11 +285,385 @@
...
+        <NODE_MONITORED>Y</NODE_MONITORED>
+        <SITENAME>Budapest Institute</SITENAME>
+        <COUNTRY_NAME>Hungary</COUNTRY_NAME>
+        <COUNTRY_CODE>HU</COUNTRY_CODE>
+        <ROC_NAME>NGI_HU</ROC_NAME>
+        <URL></URL>
+    </SERVICE_ENDPOINT>
+    <SERVICE_ENDPOINT PRIMARY_KEY="3133G0">
+        <PRIMARY_KEY>3133G0</PRIMARY_KEY>
+        <HOSTNAME>SRM.budapestinstitute-gridoperations.hu</HOSTNAME>
```
gregcorbett, line 483: perhaps change 'institute' to some synonym - here and elsewhere.

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
```diff
@@ -285,11 +285,385 @@
...
+        <NODE_MONITORED>Y</NODE_MONITORED>
+        <SITENAME>Barcs-Drava STF</SITENAME>
+        <COUNTRY_NAME>Hungary</COUNTRY_NAME>
+        <COUNTRY_CODE>HU</COUNTRY_CODE>
+        <ROC_NAME>NGI_HU</ROC_NAME>
+        <URL></URL>
+    </SERVICE_ENDPOINT>
+    <SERVICE_ENDPOINT PRIMARY_KEY="9293G0">
+        <PRIMARY_KEY>9293G0</PRIMARY_KEY>
+        <HOSTNAME>site-bdii.danubeinstitute.hu</HOSTNAME>
```
gregcorbett, line 619: perhaps change 'institute' to some synonym - here and elsewhere.

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
rowan04: ok, I'll change the sitename and hence its' services

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
rowan04: yep will change this sitename and related services too

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
rowan04: ok

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
rowan04: ok

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
```diff
@@ -285,11 +285,385 @@
...
+        <NODE_MONITORED>Y</NODE_MONITORED>
+        <SITENAME>Balaton RF</SITENAME>
+        <COUNTRY_NAME>Hungary</COUNTRY_NAME>
+        <COUNTRY_CODE>HU</COUNTRY_CODE>
+        <ROC_NAME>NGI_HU</ROC_NAME>
+        <URL></URL>
+    </SERVICE_ENDPOINT>
+    <SERVICE_ENDPOINT PRIMARY_KEY="7651G0">
+        <PRIMARY_KEY>7651G0</PRIMARY_KEY>
+        <HOSTNAME>srm.lakebalatonresearch.hu</HOSTNAME>
```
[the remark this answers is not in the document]
rowan04: ok

gregcorbett:
Looks good 😄 , the only other comment I have is - can you check the roles the Users have allign with the type of entity the role is over (see the "Role Action Map" page on GOCDB instances)?

i.e. Donna Butler has the "COD Staff" role over NGI_CH, but "COD Staff" is a Project entity role.

lib/Doctrine/deploy/sampleData/ServiceEndpoints.xml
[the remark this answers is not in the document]
rowan04: balaton is the name of the lake I made up this research facility to be next to, so i don't quite know what synonym i can give it? i was thinking of just editing the name from lakebalatonresearch.hu to something like lakebalatonpprf.hu or something, as maybe the research part is too generic. Or I could ditch balaton and call it something like lakesidePPRF. PPRF standing for particle physics research facility.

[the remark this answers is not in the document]
gregcorbett: lakesidePPRF sounds good :)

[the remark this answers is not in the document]
rowan04: got it!

rowan04:
The commit just done is to do with the site names - I will hopefully be able to fix the user roles tomorrow morning

gregcorbett:
> can you check the roles the Users have allign with the type of entity the role is over (see the "Role Action Map" page on GOCDB instances)?

I also now realise the sample User / Roles never made much sense (for example, it has always had Site level roles over NGIs), I assumed I had fixed this, but it seems that change has become trapped in #282 

It would be good to have a saner sample User / Role mapping given you are refreshing the sample data already.

The "Role Action Map" can be a little confusing (it's just confused me for a while resulting in me going down a platypus hole in the code).

Taking this as an example:

https://github.com/rowan04/gocdb/blob/8f5eb4772541a090ff04847d8022c666d21f1c16/lib/Doctrine/deploy/sampleData/UsersAndRoles.xml#L24-L28

"Site Administrator" is a Site level RoleType and should only be applied to a Site entity (i.e. the `ON_ENTITY` element should container a Site name, not an NGI name). 

Having a Site level role does grant actions over NGIs and Projects (which is how we end up with lines like this in the "Role Action Map".

image

but the "Site Administrator" only be applied to a site entity.

Does that make sense?

rowan04:
> 
When I tried to make the ON_ENTITY element a sitename, it would produce an error saying something like "no NGI's found with _sitename_". That's what i tried to do originally for roles like site administrator but had to change them all back to NGI's.

rowan04:
> > 
> 
> When I tried to make the ON_ENTITY element a sitename, it would produce an error saying something like "no NGI's found with _sitename_". That's what i tried to do originally for roles like site administrator but had to change them all back to NGI's.

Would you like me to reproduce the error?

gregcorbett:
> > > 
> > 
> > 
> > When I tried to make the ON_ENTITY element a sitename, it would produce an error saying something like "no NGI's found with _sitename_". That's what i tried to do originally for roles like site administrator but had to change them all back to NGI's.
> 
> Would you like me to reproduce the error?

That's weird, yes please 😄

rowan04:
> > > > 
> > > 
> > > 
> > > When I tried to make the ON_ENTITY element a sitename, it would produce an error saying something like "no NGI's found with _sitename_". That's what i tried to do originally for roles like site administrator but had to change them all back to NGI's.
> > 
> > 
> > Would you like me to reproduce the error?
> 
> That's weird, yes please 😄

nongis

rowan04:
And I'm heading home now. I guess we'll return to this next year :)

gregcorbett:
> I think it would be good to tackle #373 at the same time as this I think. Though we should probably just delete ServiceGroupRoles.xml and give additional servicegroup roles to the users in UsersAndRoles.xml.

Having just noticed #393 exists 🤭, lets not conflate the two - so let's not worry about servicegroups at all here :)

rowan04:
> > > > > > > 
> > > > > > 
> > > > > > 
> > > > > > When I tried to make the ON_ENTITY element a sitename, it would produce an error saying something like "no NGI's found with _sitename_". That's what i tried to do originally for roles like site administrator but had to change them all back to NGI's.
> > > > > 
> > > > > 
> > > > > Would you like me to reproduce the error?
> > > > 
> > > > 
> > > > That's weird, yes please 😄
> > > 
> > > 
> > > nongis
> > 
> > 
> > this was the code torch
> 
> ah, looks like `<ENTITY_TYPE>group</ENTITY_TYPE>` is the culprit. Going a bit deeper it looks like UsersAndRoles.xml is based on an older version of the API output from GOCDB's `get_user` method 😢
> 
> Looking at the output now, `<ENTITY_TYPE>` is either `site`, `ngi`, `project`, `servicegroup`.
> 
> * AddSiteRoles.php looks like it already looks for `<ENTITY_TYPE>` equalling `site`: https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/deploy/AddSiteRoles.php#L22
> * AddGroupRoles.php should probably be renamed AddNGIRoles.php and only focus on `<ENTITY_TYPE>` equalling `ngi`: https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/deploy/AddGroupRoles.php#L17
> * AddEgiRoles.php should probably be generalised into AddProjectRoles.php and focus on and only focus on `<ENTITY_TYPE>` equalling `project` (and not assume it's working on the EGI project) see https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/deploy/AddEgiRoles.php#L52 and https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/deploy/AddEgiRoles.php#L57
> 
> I'm not sure what's going on with `<ENTITY_TYPE>` equalling `servicegroup`. We don't seem to have a AddServiceGroupRoles.php equivalent, but we do have some sample ServiceGroupRoles.. I think it would be good to tackle #373 at the same time as this I think. Though we should probably just delete ServiceGroupRoles.xml and give additional servicegroup roles to the users in UsersAndRoles.xml.

I see (I think!)

rowan04:
> > I think it would be good to tackle #373 at the same time as this I think. Though we should probably just delete ServiceGroupRoles.xml and give additional servicegroup roles to the users in UsersAndRoles.xml.
> 
> Having just noticed #393 exists 🤭, lets not conflate the two - so let's not worry about servicegroups at all here :)

Ok.

lib/Doctrine/deploy/AddProjectRoles.php
```diff
@@ -91,9 +88,13 @@
...
         }
 
-        $doctrineRole = new Role($roleType, $doctrineUser, $egi, 'STATUS_GRANTED');
-        $entityManager->persist($doctrineRole);
+        if ((string) $role->ON_ENTITY == "EGI") {
+            $doctrineRole = new Role($roleType, $doctrineUser, $egi, 'STATUS_GRANTED');
+            $entityManager->persist($doctrineRole);
+        } else {
+            continue;
+        }
```
gregcorbett, line 96: Can this be a look up on the underlying project entity and, regardless of whether it's EGI or another project, add the appropriate role?

Similar to https://github.com/rowan04/gocdb/blob/384-refresh_sample_data/lib/Doctrine/deploy/AddNGIRoles.php#L65-L78

gregcorbett: and ammend the docstring at the top of the file accordingly.

lib/Doctrine/deploy/AddProjectRoles.php
```diff
@@ -5,33 +5,35 @@
 
 /**
  * AddEgiRoles.php: Loads a list of roles from the get_user PI
- * query output (XML), finds the EGI roles and inserts them into
- * the doctrine prototype.
+ * query output (XML), finds whether they are EGI roles or not
+ * and if they are, adds them to the EGI project and 
+ * insert them into the doctrine prototype.
+ * If they are not an EGI role, they are not added to a project
  * XML format is the output from get_roc_list PI query, e.g:
```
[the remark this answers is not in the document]
rowan04: ok

rowan04:
I found that the INSTALL.md information about the sample data is now incorrect: https://github.com/GOCDB/gocdb/blob/f6d4a1b750eecc12c2ca4aa4f38a496154f27b29/INSTALL.md#optional-deploy-sample-data
However it will need to be edited for the remove simple sample data (#415) issue as it also contains information about the simple sample data, so should this file be left for that pull request or should I edit the bit about the normal sample data here?

gregcorbett:
> I found that the INSTALL.md information about the sample data is now incorrect:

Good spot :) 

> should I edit the bit about the normal sample data here?

Yes please, else we run the risk of forgetting about it :)

gregcorbett:
This is looking great 😄, one final thing - cna you remove any `CIC Staff` roles from UsersAndRoles.xml. That is a legacy role that we no longer use.

rowan04:
> This is looking great 😄, one final thing - cna you remove any `CIC Staff` roles from UsersAndRoles.xml. That is a legacy role that we no longer use.

done

gregcorbett (approved):
LGTM 😄

gregcorbett:
Rebasing

lib/Doctrine/deploy/AddNGIs.php
rowan04: done in #428

PR #396, opened 2022-11-14
Allow single quote in user identifiers to master

gregcorbett:
GOCDB has historically allowed single quote characters in a users DN (they should not be in an IGTF DN, but are not expressly forbidden).

However the new user identifiers mechanism does not currently support single quote characters. This means that users with a single quote in their DN cannot:
- access GOCDB (they can't be migrated to the new user identifiers mechanism), if they are already registered
- cannot register with GOCDB, if they are not already.

Indeed, some user DNs in GOCDB do contain single quote characters and one user cannot recover their account because of this issue.

Subject to a positive review here, I'll invite the user who raised this issue to try and recover their account again. If that is successful, I'll bump the version number up to 5.10.1 and merge this in. This PR deliberately merges into `master` so that it can be released in isolation.

### DIFF
--- a/htdocs/web_portal/static_html/goc5_logo.html
+++ b/htdocs/web_portal/static_html/goc5_logo.html
@@ -4,7 +4,7 @@
 <!--	<img src="img/Logo-1.6.png" class="logo_image" height="39" style="vertical-align: middle;"/>-->
   <h3    class="Logo_Text Small_Bottom_Margin Standard_Padding"
          style="vertical-align: middle; margin-left: 0.2em;">
-         GOCDB 5.10.0
+         GOCDB 5.10.1
      </h3>
 
 </a>


config/gocdb_schema.xml
```diff
@@ -493,7 +493,7 @@
 	    <!-- negation regex - allow all chars except negated -->
```
ineilson, line 493:
```suggestion
```

config/gocdb_schema.xml
```diff
@@ -493,7 +493,7 @@
...
         <!--Allows most characters, including whitespace. However, no trailing or leading whitespace permitted -->
         <!-- <regex>/^[^`'\"&lt;&gt;\s]+(\s*[^`'\"&lt;&gt;\s]+)*$/</regex> -->
-        <regex>/^[^`'\"&lt;&gt;\s]+(\s*[^`'\"&lt;&gt;\s]+)*$/</regex>
+        <regex>/^[^`\"&lt;&gt;\s]+(\s*[^`\"&lt;&gt;\s]+)*$/</regex>
```
ineilson, line 496: This now seems to be the same expression as is commented out on the line before. Do we know why the single quote was, presumably, deliberately removed?

gregcorbett: > This now seems to be the same expression as is commented out on the line before

Good spot!

> Do we know why the single quote was, presumably, deliberately removed?

Alas no, this was the commit that introduced it - https://github.com/GOCDB/gocdb/commit/037e16cad456d2df2e334e55b576c6137665e52d

config/gocdb_schema.xml
gregcorbett: Hmmm 🤔

gregcorbett:
Annoyingly I can't make this a GitHub suggestion, but taking onboard both you comments @ineilson - how about the following instead as the VALUE definition of a User Identifier?

```
<field>
    <!-- Identifier key value -->
    <fname>VALUE</fname>
    <length>255</length>
    <!-- Allows most characters, including whitespace.
         However, no trailing or leading whitespace is permitted.
      -->
    <regex>/^[^`'\"&lt;&gt;\s]+(\s*[^`'\"&lt;&gt;\s]+)*$/</regex>
</field>
```

config/gocdb_schema.xml
gregcorbett: See https://github.com/GOCDB/gocdb/commit/83f38c4177c054ca30e85e308c4bb7ef4240bb39

config/gocdb_schema.xml
gregcorbett: See https://github.com/GOCDB/gocdb/commit/83f38c4177c054ca30e85e308c4bb7ef4240bb39

ineilson (approved):

PR #398, opened 2022-11-15
Replace tabs for spaces and reindent gocdb_schema.xml

gregcorbett:
Whislt working on #396 and #397, I noticed this file used a mix of spaces and tabs. It was also indented inconsistently.

This PR fixes both.

### DIFF
--- a/config/gocdb_schema.xml
+++ b/config/gocdb_schema.xml
@@ -2,601 +2,601 @@
 <schema>
 
     <!--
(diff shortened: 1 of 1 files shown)


gregcorbett:
[Viewing without whitespace changes](https://github.com/GOCDB/gocdb/pull/398/files?diff=split&w=1) may be useful here.

ineilson (approved):

PR #399, opened 2022-11-15
Remove "IRIS IAM" login button from landing page to master

gregcorbett:
GOCDB doesn't currently support IRIS IAM "out of the box", so the button doesn't work in most deployments and it does cause confusion for users.

Subject to a positive review here, I'll bump the version number up (probably to 5.10.~~2~~3) and merge this in. This PR deliberately merges into master so that it can be released in isolation.


gregcorbett:
I'd rather role this out as part of [milestone#13](https://github.com/GOCDB/gocdb/milestone/13). As such I will close this in favour of #432.

PR #401, opened 2022-11-29
Add an access_test method to API to master

gregcorbett:
In preparation for restricting access to personal data, this new method enables Service Owners to be able to verify their hosts will retain access to the API after access to personal data is restricted.

Required a change to `authByIdentifier` and `isRestrictPDByRole` to force the restrictions for the test method regardless of the wider API access rules. I haven't included Users in this API check as I don't want to encourage human users talking to the API.

Subject to a positive review here, I'll bump the version number up (probably to 5.10.2) and merge this in. This PR deliberately merges into master so that it can be released in isolation.

It's running in preproduction: https://gocdb-preprod.egi.eu

### DIFF
--- a/htdocs/PI/index.php
+++ b/htdocs/PI/index.php
@@ -141,6 +141,10 @@ class PIRequest {
             $em = \Factory::getEntityManager();
 
             switch ($this->method) {
+                case "access_test":
+                    $this->authByIdentifier(true);
+                    return "<success/>\n";
+                break;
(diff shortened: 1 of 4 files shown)


gregcorbett:
@ineilson is going to look at addressing the codeclimate issues in a seperate PR, as well as look at returning valid XML in the result of a failure in an upcoming PR 😄

PR #402, opened 2022-11-29
Add an access_test method to API to dev

gregcorbett:
Subject to #401 being reviewed, this PR contains no unreviewed code.


ineilson:
If xml is returned, shouldn't it be roughly complete? Something like 
```
<?xml version="1.0" encoding="UTF-8"?>
<results identity="/C=UK/O=eScience/OU=CLRC/L=RAL/CN=my name">
<authorized>true</authorized>
</results>
```
maybe, consistent with the other method returns? Lazy parsing of the result could still just grep for `<authorized>true` (or success, whatever) but there is a chance of returning at least the DN/id or other 'useful' diagnostics?

gregcorbett:
Sounds good, please do suggest the code you put together that generated that XML

ineilson (approved):

PR #409, opened 2022-12-07
Functionality for warning email and API credential delete

ineilson:


gregcorbett:
Looks great 😄 , is the idea that it would be used as (for example):

```
# delete API creds not used in 6 months
php ManageAPICredentials -d -t 6
# warn users about API creds not used in 6 months
php ManageAPICredentials -w -t 3
```

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,239 @@
...
+    $userEmail = $user->getEmail();
+    $siteName = $api->getParentSite()->getShortName();
+    $siteEmail = $api->getParentSite()->getEmail();
+
+    $headers = "From: GOCDB <gocdb-admins@mailman.egi.eu>";
+    $headers .= "\r\nCC: $siteName <$siteEmail>";
+    $subject = "GOCDB: Site API credential deletion notice";
+
+    $body = "Dear " . $user->getForename() . ",\n\n" .
+        "The API credential associated with the following identifiers " .
```
gregcorbett, line 127: API credentials associated to accounts, not identifers, right?

Maybe here it's better to to say ""The API credential added by user X registered at site ..."

gregcorbett: and then remove `$body .= "Identifier: " . $api->getIdentifier() . "\n";`

gregcorbett: ah sorry, I'm being dense - it's the API credential identifier you're displaying 🤦 

Maybe it would be good to still say which user added it? So if you are actioning the email because of the cc you have a bit more context?

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,239 @@
...
+    {
+        if ($message != '') {
+            print
+            (
+                "Error: $message\n"
+            );
+        }
+        print
+        (
+            "Usage: php ManageAPICredentials [--help | -h] [--dry-run] [[--warn | -w] | [--delete | -d ]]\n" .
```
gregcorbett, line 211:
```suggestion
            "Usage: php ManageAPICredentials.php [--help | -h] [--dry-run] [[--warn | -w] | [--delete | -d ]]\n" .
```

gregcorbett:
Should this script handle the regular renewal of used api crednetials (i.e. this credential is being used, but we need a user to confirm it is actually needed every `Z` months) as well?

Or would that be better as a seperate script do you think?

ineilson:
> Looks great 😄 , is the idea that it would be used as (for example):
> 
> ```
> # delete API creds not used in 6 months
> php ManageAPICredentials -d -t 6
> # warn users about API creds not used in 6 months
> php ManageAPICredentials -w -t 3
> ```

For discussion. I had a version with two thresholds, but the logic wasn't right. I think there's possibly a place for second parameter input to the warning option to say "it'll be deleted when it gets to x months unused"
``` 
php ManageAPICredentials -d -t 6
# warn users about API creds not used in 6 months
php ManageAPICredentials -w -p 5 -t 6
```
So, warn at 5 for 6 month deletion. Or similar?

resources/ManageAPICredentials.php
ineilson: OK will add user email for context.

ineilson:
> Should this script handle the regular renewal of used api crednetials (i.e. this credential is being used, but we need a user to confirm it is actually needed every `Z` months) as well?
> 
> Or would that be better as a seperate script do you think?

First thought would be to put all API Cred management in one place as easier for the user/admin, even though internally the options might not overlap too much.

It feels like this confirming ownership periodically might be tied in with the management of inactive users - warning that this credential will be deleted, even though it's being used, because the owning use hasn't logged in, and suggesting taking new ownership will fix it?

gregcorbett:
> It feels like this confirming ownership periodically might be tied in with the management of inactive users - warning that this credential will be deleted, even though it's being used, because the owning use hasn't logged in, and suggesting taking new ownership will fix it?

Hmmmm I worry tying it up with inactive users wouldn't capture the case where a user is active and the api credential is in use - but the credential isn't actually needed.

For reference, I'm imagining API cred clean up would happen at a higher rater than user clean up (lets say delete after 6 months of inactivty. and periodically confirm every 12?). And API cred clean up could be done first (before deletion of inactive users). 

That way, I don't see an inactive user owning any used api creds, they would always get deleted first unless the user becomes active and touches them (or someone else takes the API cred).

ineilson:
> Hmmmm I worry tying it up with inactive users wouldn't capture the case where a user is active and the api credential is in use - but the credential isn't actually needed.

🤔 Used but not needed: meaning that, kind of from the site's perspective, somebody has set up a useless periodic API operation, or they registered a personal credential for the API and subsequently stopped using it? This probably needs a state diagram!

gregcorbett:
>Used but not needed: meaning that, kind of from the site's perspective, somebody has set up a useless periodic API operation

Exactly

> This probably needs a state diagram!

Maybe a nice pre christmas activity for me 😋

ineilson:
> > For discussion. I had a version with two thresholds, but the logic wasn't right. I think there's possibly a place for second parameter input to the warning option to say "it'll be deleted when it gets to x months unused"
> 
> For user deletion, a single run of the script will both warn and delete users based on the thresholds set in the `warning_threshold` and `deletion_threshold` options. Maybe we could do that here? Then we only need to run the script once rather than twice (once to warn users and once to delete them).

So it sounds like options for the two thresholds you cite, where deletion_threshold must be > warning_threshold. If either isn't specified, the missing op is not performed, but at least one must be provided. i.e. no threshold default values. The default action is to execute all the ops specified unless the --dry-run option is given.

Okay?

The only 'problem' I see is that warnings (for the offending credentials) will be sent out each time the script is run, so if it is, say, run monthly, 'countdown' warning emails will be sent as the time approaches the deletion threshold. Probably that's ok, but only provided it's not run daily!

gregcorbett:
> Okay?
> 
> The only 'problem' I see is that warnings (for the offending credentials) will be sent out each time the script is run, so if it is, say, run monthly, 'countdown' warning emails will be sent as the time approaches the deletion threshold. Probably that's ok, but only provided it's not run daily!

Sounds good, I imagine the script will run O("Monthly") rather than O("Daily").

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,305 @@
...
+
+    if ($options->isShowHelp()) {
+        return;
+    }
+
+    $baseTime = new DateTime("now", new DateTimeZone('UTC'));
+
+    $deletedCreds = [];
+
+    if (!is_null($options->getDelete())) {
```
gregcorbett, line 58: I wonder if this would be clearer if the `!is_null($options->getDelete())` was contained within the ManageAPICredentialsOptions class, and here have something like `if ($options->willDeleteCreds()` or a better named method :p

gregcorbett: and similar for the warning variant

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,305 @@
...
+    if (!is_null($options->getWarn())) {
+        $warnedCreds = warnUsers(
+            $options->isDryRun(),
+            $entityManager,
+            $baseTime,
+            $options->getWarn(),
+            $options->getDelete(),
+            $fromEmail,
+            $replyToEmail,
+            $deletedCreds
```
gregcorbett, line 76: why does the warning method need to know the deleted credentials?

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,305 @@
...
+    $deletionThreshold,
+    $fromEmail,
+    $replyToEmail,
+    $deletedCreds
+) {
+    $creds = getCreds($entityManager, $baseTime, $warningThreshold);
+
+    if ($dryRun) {
+        // Remove the credentials that would have been deleted if not in dry-run mode
+        $creds = array_udiff($creds, $deletedCreds, '\gocdb\scripts\compareCredIds');
```
gregcorbett, line 167: why might there be duplicates in credentials list?

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,305 @@
...
+ * instead of sending emails.
+ *
+ * @param bool          $dryRun             If true no action is taken and a report is generated instead
+ * @param \Doctrine\Orm\EntityManager $entitymanager A valid Doctrine Entity Manager
+ * @param \DateTime     $baseTime           Time from which interval of no-use is measured
+ * @param int           $warningThreshold   The number of months of no-use which triggers warning emails
+ * @param int           $deleteThreshold    The number of months of no-use which will trigger deletion
+ * @param string        $fromEmail          Email address to use as sender's (From:) address
+ * @param string        $replyToEmail       Email address for replies (Reply-To:)
+ * @param array         $deletedCreds       An array of credentials which have been selected for deletion
```
gregcorbett, line 148: maybe `$deletedCreds` would be better documented as $ignoreCreds?

resources/ManageAPICredentials.php
gregcorbett: never mind, have read on 😋

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,305 @@
...
+
+    // Some code useful for testing
+    /*
+    print("\n!!!DUMMY SITE EMAIL!!! was $siteEmail\n\n");
+    $siteEmail = 'Name <name@address.com>';
+    */
+    /*
+    print("\n!!!DUMMY USER EMAIL!!! was $userEmail\n\n");
+    $userEmail = 'AnotherName <anothername@address.com>';
+    */
```
gregcorbett, line 249: I think it would be better for testing to use a dummy data set - so lets remove this commented out code?

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,305 @@
...
+                           "Cc: $siteEmail");
+    if (strlen($replyToEmail) > 0 && $fromEmail !== $replyToEmail) {
+        $headersArray[] = "Reply-To: $replyToEmail";
+    }
+    $headers = join("\r\n", $headersArray);
+
+    $subject = "GOCDB: Site API credential deletion notice";
+
+    $body = "Dear " . $user->getForename() . ",\n\n" .
+        "The API credential associated with the following identifiers registered\n" .
```
gregcorbett, line 261:
```suggestion
        "The API credential associated with the following identifier registered\n" .
```

The API credential will only be associated with one identifier, right?

gregcorbett:
I presume the change to `htdocs/PI/index.php` was needed to catch the case where the identity string of a user is also a API credential?

gregcorbett:
For readability of `resources/ManageAPICredentialsOptions.php`, could we have a one line space between functions?

i.e.

```
    public function getWarn()
    {
        return $this->warn;
    }

    public function getDelete()
    {
        return $this->delete;
    }

```

resources/ManageAPICredentials.php
```diff
@@ -0,0 +1,305 @@
...
+try {
+    $options = new ManageAPICredentialsOptions();
+
+    if ($options->isShowHelp()) {
+        return;
+    }
+
+    $baseTime = new DateTime("now", new DateTimeZone('UTC'));
+
+    $deletedCreds = [];
```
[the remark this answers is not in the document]
ineilson: warnedCreds is not actually used, but gives some symetry to the two methods, but, I agree, it's better if it is initialised.

resources/ManageAPICredentials.php
ineilson: Good plan.

resources/ManageAPICredentials.php
ineilson: But it's messy. I propose to rearrange using a single up-front fetched list from which creds are popped after deletion.

resources/ManageAPICredentials.php
ineilson: Probably, but as suggested in comment above, to use a single list from which deleted (even in dry-run) creds are removed is cleaner.

resources/ManageAPICredentials.php
ineilson: Because the list is fetched twice - once for delete and once for warning - I propose to use a single list so duplication check would be unnecessary.

resources/ManageAPICredentials.php
ineilson: OK

resources/ManageAPICredentials.php
ineilson: Well spotted.

gregcorbett:
I assume this has been closed while work on a test case is ongoing?

PR #412, opened 2022-12-19
GT-124 Make references to operational tools configurable

rowan04:
- added config options for helpdesk link, helpdesk support unit, request tracker and community documentation
- so that different instances of gocdb can have links point directly to their tooling, instead of having general references


rowan04:
Note - this currently doesn't work and the gocdb webpage doesn't have the correct links. I am still working on this

htdocs/web_portal/index.php
```diff
@@ -208,6 +208,11 @@ function Draw_Page($Page_Type) {
             //rejectIfNotAuthenticated();
             Draw_Static_HTML();
             break;
+        case "Static_PHP":
+            //rejectIfNotAuthenticated();
+	    //require_once __DIR__.'/components/Draw_Components/draw_page_components.php';
```
rowan04, line 213: This line is to be removed

htdocs/web_portal/index.php
```diff
@@ -686,6 +712,25 @@ function Get_Static_Page_Contents($Page_Name) {
...
+/* Get the contents of the static PHP page specified in $PHP_Page_Name
+ * if the page name isn't specified then return a blank string */
+function Get_Static_PHP_Page_Contents($PHP_Page_Name) {
+    require_once __DIR__.'/components/Draw_Components/draw_page_components.php';
+    $phpDir = __DIR__."/static_php";
+    $Available_Static_Pages = Get_Directory_Contents($phpDir);
+    if(!isset($Available_Static_Pages[$PHP_Page_Name])) {
+        return "";
+    }
+    $PHP = Get_File_Contents($phpDir."/".$PHP_Page_Name);
```
rowan04, line 724: Here's the problem (line 724) - Get_File_Contents just gets the contents of the file, it doesn't execute it. Therefore the php parts aren't being executed, just written out. idk what to replace it with though

rowan04: Fixed this - I just used the commented out statements below I'd tried before! I had to use the `systemctl reload httpd` to update it, which is maybe why it hadn't worked before.

gregcorbett:
Hmmmmm, I wonder if moving the Help_And_Contact to be fully within the view/controller setup would be a cleaner solution than using `ob_start()` et al.

my_sites.php is a relatively small example of the view/controller setup in use:
- https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/controllers/my_sites.php
- https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/views/my_sites.php

So for this, you'd have to create, for example:
- `htdocs/web_portal/controllers/help_and_contact.php`. 
- `htdocs/web_portal/views/help_and_contact.php`. 

The controller is where the code in htdocs/web_portal/static_php/operationalLinks.php would go, except with the output contained in a `params` array.

The view is where the contents of `htdocs/web_portal/static_php/Help_And_Contact.php` would go, except with references to the `params` array, i.e. `params['HelpdeskLink']` instead of `$helpdeskLink`.

You'd have to add an additonal case for PageType=Help_And_Contact in [web_portal/index.php](https://github.com/GOCDB/gocdb/blob/dev/htdocs/), similar to the [My_Sites case](https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/index.php#L300), but it would mean less new code in index.php.

Could you give that a go?

rowan04:
> Hmmmmm, I wonder if moving the Help_And_Contact to be fully within the view/controller setup would be a cleaner solution than using `ob_start()` et al.
> 
> my_sites.php is a relatively small example of the view/controller setup in use:
> 
> * https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/controllers/my_sites.php
> * https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/views/my_sites.php
> 
> So for this, you'd have to create, for example:
> 
> * `htdocs/web_portal/controllers/help_and_contact.php`.
> * `htdocs/web_portal/views/help_and_contact.php`.
> 
> The controller is where the code in htdocs/web_portal/static_php/operationalLinks.php would go, except with the output contained in a `params` array.
> 
> The view is where the contents of `htdocs/web_portal/static_php/Help_And_Contact.php` would go, except with references to the `params` array, i.e. `params['HelpdeskLink']` instead of `$helpdeskLink`.
> 
> You'd have to add an additonal case for PageType=Help_And_Contact in [web_portal/index.php](https://github.com/GOCDB/gocdb/blob/dev/htdocs/), similar to the [My_Sites case](https://github.com/GOCDB/gocdb/blob/dev/htdocs/web_portal/index.php#L300), but it would mean less new code in index.php.
> 
> Could you give that a go?

ok

gregcorbett (changes requested):
see my previous comment

gregcorbett (approved):
This looks great :). I want to prepare the necassary config change, in a sandbox, at the same time I merge this.

gregcorbett:
rebasing to fix conflict

PR #417, opened 2023-01-17
Same API identifier at multiple sites to master.

ineilson:
Allow the same identifier to be used for APIauthentication entities at multiple sites

PR #418, opened 2023-01-17
Same API identifier at multiple sites to dev.

ineilson:


gregcorbett (approved):
All code reviewed in #417

PR #420, opened 2023-01-23
Cast all returns from Config.php to appropriate types

### LINKED ISSUE #413: SimpleXML objects being returned from Config.php, not strings
There is potential for confusion or error (largely created by me) in that this statement and other function returns from the Config class  - 

https://github.com/GOCDB/gocdb/blob/f94e9807592ab18ee4318758c94329d5e6b95bf4/lib/Gocdb_Services/Config.php#L541

When $bannerText is subsequently returned from the function, it actually returns a SimpleXMLElement object and **not** a string. Subsequent use of the returned object usually works in context, due to implicit casting to a string, but it can behave unexpectedly e.g. '==' comparison between two such returned values from different elements will fail even when the string values are the same.

It would be safer if all returns from Config.php were explicitly cast to strings or other type as and where appropriate.

rowan04:
- as per issue request, all returns from config.php are now explicitly cast to appropriate types
- this is to prevent confusion and error as before returns were being cast as a SimpleXMLElement object which can cause unexpected behaviour
- new lines were also added in and around some of the functions where they were missing to make the text more readable

Resolves #413

### DIFF
--- a/lib/Gocdb_Services/Config.php
+++ b/lib/Gocdb_Services/Config.php
@@ -267,57 +267,70 @@ class Config
      */
     public function IsPortalReadOnly()
     {
-        $localInfo = $this->GetLocalInfoXML();
-        if (strtolower($localInfo->read_only) == 'true') {
+        $portalReadOnly = (string) $this->GetLocalInfoXML()->read_only;
+        if (strtolower($portalReadOnly) == 'true') {
             return true;
         }
 
         return false;
     }
+
     /**
      * returns the url of the Acceptable Use Policy for display on the landing page
      * @return string
      */
     public function getAUP()
     {
-        return  $this->GetLocalInfoXML()->aup;
+        $aup = (string) $this->GetLocalInfoXML()->aup;
+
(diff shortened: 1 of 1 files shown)


tofu-rocketry:
@gregcorbett I guess we don't have an existing unit test that could be extended to cover this?

gregcorbett:
As discussed, @rowan04 would you be able to take a look at implementing a unit test to check the public methods in `lib/Gocdb_Services/Config.php` return the correct type and value for some given input?

rowan04:
The force push was to amend a commit name/message

rowan04:
After doing some tests on some of the functions in Config.php, this type casting method works. However, there is still some work for me to do to create some proper unit tests

gregcorbett (changes requested):
Unit test writing that is in progress.

rowan04:
Unfortunately, due to me starting a new rotation, I am unable to write the unit tests for this pr. See #486

rowan04:
fixed conflicts

PR #429, opened 2023-02-08
Remove type from uniqueness check

ineilson:
The identifier of an API credential is assumed to be unique irrespective of the type, so the type need not be
included in the uniqueness check. This also allows transition from legacy type string "X509" to "X.509" without
throwing an error when a credential user is reassigned.


gregcorbett (approved):
LGTM 😄

PR #430, opened 2023-02-17
GT-56: Add base class for OIDC AuthTokens, add EOSCAAI AuthToken

gregcorbett:
This PR:

- adds an abstract class that future OIDC based IDP integrations can extended.
- provides an concrete class for integrating the EOSC (demo) IDP via OIDC

Rather than proliferate `die()` calls, the OIDC abstract class throws exceptions which are now handled (granted this handling then calls `die()` because it can't make use of `show_view('error.php',...`)

This code is running on [gocdb-preprod.eosc-portal.eu](https://gocdb-preprod.eosc-portal.eu).

Supersedes #348.

PR #431, opened 2023-02-17
Add a check on the issuer of the token for IRIS IAM

gregcorbett:
Apache is not already doing this for us

Either this needs to go in [milestone 13](https://github.com/GOCDB/gocdb/milestone/13), or a PR that refactors the IRIS IAM token to use the base class introduced by #430 (which also includes a check on the issuer.)


gregcorbett:
> Either this needs to go in [milestone 13](https://github.com/GOCDB/gocdb/milestone/13), or a PR that refactors the IRIS IAM token to use the base class introduced by https://github.com/GOCDB/gocdb/pull/430 (which also includes a check on the issuer.)

This is only really an issue when we have multiple OIDC integrated IdP supported in the main codebase, so bumping to [milestone 14](https://github.com/GOCDB/gocdb/milestone/14).

PR #432, opened 2023-02-23
Remove "IRIS IAM" login button from landing page

gregcorbett:
GOCDB doesn't currently support IRIS IAM "out of the box", so the button doesn't work anyway.


rowan04 (approved):
LGTM :)

PR #435, opened 2023-03-06
Fix incorrect function call in config

### LINKED ISSUE #434: Incorrect function call in Config
https://github.com/GOCDB/gocdb/blob/69a6e01a48aba5543a646ed481fb2cd6c8f6a69f/lib/Gocdb_Services/Config.php#L555

As I only recently wrote this code and was recently reviewed and merged in #411, I have no idea how this wasn't noticed.
Should be `GetLocalInfoXML()` with a capital L.

rowan04:
- there was a typo in Config.php when calling GetLocalInfoXML()
- this fixes it

Resolves #434

### DIFF
--- a/lib/Gocdb_Services/Config.php
+++ b/lib/Gocdb_Services/Config.php
@@ -552,7 +552,7 @@ class Config
 
     public function getEmailTo()
     {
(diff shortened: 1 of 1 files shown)


gregcorbett (approved):
👍

PR #437, opened 2023-03-07
GT-91 Manage API credentials

ineilson:
Script and accompanying tests to filter API credentials by last-use time to allow sending warning emails and deletion of unused credentials.

Old PR: https://github.com/GOCDB/gocdb/pull/409

### DIFF
--- a/htdocs/PI/index.php
+++ b/htdocs/PI/index.php
@@ -404,6 +404,16 @@ class PIRequest
         $admin = false;
         $authenticated = false;
 
+        // Check if incoming identifier is a registered API Authentication credential.
+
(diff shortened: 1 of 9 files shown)


gregcorbett (approved):
LGTM

gregcorbett:
This only deals with _unused_ API credentials, right?

It doesn't deal with the regular renew of API credentials right? i.e. this won't catch the case where somebody has set up a useless periodic API operation, and subsequently stopped needing it, but the access continues.

ineilson:
> This only deals with _unused_ API credentials, right?

Correct. It's all keyed on last-used time.

PR #442, opened 2023-03-22
TEST: Use the beta channel phpcodesniffer

gregcorbett:

PR #441, opened 2023-03-21
Use the beta channel phpcodesniffer

gregcorbett:
Needed to enforce PSR12, see: https://docs.codeclimate.com/docs/phpcodesniffer#standard

PR #443, opened 2023-03-22
Use the beta channel phpcodesniffer

gregcorbett:
Needed to enforce PSR12, see: https://docs.codeclimate.com/docs/phpcodesniffer#standard

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,6 +2,9 @@ version: "2"
 plugins:
   phpcodesniffer:
     enabled: true
+    # Need the beta channel to enforce PSR12,
+    # see https://docs.codeclimate.com/docs/phpcodesniffer#standard.
+    channel: "beta"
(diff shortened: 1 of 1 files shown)


gregcorbett:
>14841 issues to fix?!

Haha yep, I hoping that means codeclimate is now picking up PSR12 style issues, and it's treating them as new issues because these issues aren't currently detected on `dev`. But I can't confirm that because there are too many issues for codeclimate to render (which is why I had some test PRs: #441 and #442).

Another hope is that PRs checked after this one has been merged will be compared to this new 14841 baseline - and as such the change will be useful.

If merging this PR means that all future PRs codeclimate just says "too many issues", then this change is not useful and I shall revert.

ineilson (approved):
Looks fine with proviso I'm not sure exactly what the rules governing beta channel release are.

PR #454, opened 2023-05-09
Update codebase to PHP 7.4

gregcorbett:
Replaces PR #357 as fork has been deleted.

Also integrates fixes for misc PHP 7.4 notices and warnings.

PR #357, opened 2022-06-16
Update to ORM 2.5 and PHP 7.4

ineilson:
PR as requested - https://github.com/GOCDB/gocdb/pull/344#issuecomment-1122620062

Also integrates fixes for misc PHP 7.4 notices and warnings https://github.com/GOCDB/gocdb/pull/344/commits/4b942767a19276ba7dd6abd7f25e362a185ee828

PR #455, opened 2023-05-09
Update ORM to 2.11

### LINKED ISSUE #335: Move to newer version of Doctrine (2.5 at least)
The version of Doctrine we currently use is no longer supported (https://www.doctrine-project.org/projects/orm.html). We were stuck on it because the next version (2.5) required php 5.4  or greater. With #238 we removed php 5.3 support and so can now migrate upwards. This may be true for other components as well.

Pretty careful testing would be needed

gregcorbett:
Replaces closed #344 as fork has been deleted.

Depends on #454

Closes #335

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "stfc-egi/gocdb",
     "description": "A web portal and REST style API for e-Infrastructure topology managment",
     "require": {
-	"doctrine/orm": "2.7.*"
+	"doctrine/orm": "2.8.*"
     },
     "require-dev": {
         "phpunit/phpunit": "4.8.*",
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "stfc-egi/gocdb",
     "description": "A web portal and REST style API for e-Infrastructure topology managment",
     "require": {
-	"doctrine/orm": "2.10.*",
+	"doctrine/orm": "2.11.*",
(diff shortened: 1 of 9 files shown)


rowan04 (commented):
The changes in every file other than composer.lock are the same as in PR #344.
There is a difference of about 300 additions/deletions between the changes made to composer.lock in this PR and #344, with more changes being made to #344.
There are a couple more lines changed to composer.json in #344 than this PR - I think this is due to updates in other PRs made between these two PRs, and not a problem.
Apart from that, the other files are the same.

rowan04:
> > The changes in every file other than composer.lock are the same as in PR #357. There is a difference of about 300 additions/deletions between the changes made to composer.lock in this PR and #344, with more changes being made to #344. There are a couple more lines changed to composer.json in #344 than this PR - I think this is due to updates in other PRs made between these two PRs, and not a problem. Apart from that, the other files are the same.
> 
> I assume the first reference to #357 should be #344 ?
> 
> Thanks again for this, RE: composer.json - yes, the difference seems to be because [4dad3cd](https://github.com/GOCDB/gocdb/commit/4dad3cd7a02c7ac42ffbaa9f3cde0f0a31a1a314) has already been merged.
> 
> As mentioned in #454, composer.lock needs regenerating anyway before merging - so if that's the only hard to explain difference I am happy with that 😋

yep i've updated it to #344.

PR #459, opened 2023-06-28
[GT-180] Add API credentials support to search functionality

### LINKED ISSUE #452: Expand "Search" functionality in portal to cover API credentials
It's often quite useful to be able to identify a site assosicated with an API credential, and at the moment I have to log into the database to do so, which is less than ideal.

Perhaps the "Search" functionality in portal could return the owning Site?

Sae126V:
It's often quite useful to be able to identify a site associated with an API credential, and at the moment I have to log into the database to do so, which is less than ideal.

Perhaps the "Search" functionality in portal could return the owning Site?

"Resolves #452 "

### DIFF
--- a/htdocs/web_portal/controllers/search.php
+++ b/htdocs/web_portal/controllers/search.php
@@ -23,7 +23,7 @@
 function search() {
     require_once __DIR__.'/../../../lib/Gocdb_Services/Factory.php';
 
-
+    $params = [];
     $dn = Get_User_Principle();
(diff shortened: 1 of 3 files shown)


gregcorbett:
Looks good to me, going to rebase this on top of the latest dev as it is out of date (this should also re-trigger the codeclimate analysis as well, which seems to have aged out of being available).

gregcorbett (approved):

PR #460, opened 2023-07-07
[GT-182] Add restrictions on who can create SG's

### LINKED ISSUE #364: restrict who can create ServiceGroups
Currently, any authenticated (and registered) user can create a ServiceGroup. This feels too open.

Not sure who exactly should be able to create a ServiceGroup, maybe restrict it to Users with a role over an entity in GOCDB. That would atleast mean a human has said the user has some business doing things with GOCDB.

Sae126V:
If a user is an `Admin` -> they are allowed to ADD new service group's.

If a user is NOT an `Admin`:
1) Case1: User is NOT allowed to add new SG's. If a User have no roles assigned.
2) Case2: User is allowed to add new SG's. If a user is registered and have at least one role assigned with 'STATUS_GRANTED'.

Resolves #364

### DIFF
--- a/htdocs/web_portal/controllers/service_group/add_service_group.php
+++ b/htdocs/web_portal/controllers/service_group/add_service_group.php
@@ -22,6 +22,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
   /*====================================================== */
+use Exception;
+
 require_once __DIR__ . '/../../../../lib/Gocdb_Services/Factory.php';
 require_once __DIR__ . '/../../components/Get_User_Principle.php';
 require_once __DIR__ . '/../utils.php';
@@ -72,6 +74,20 @@ function draw($user) {
         // If the user is registered they're allowed to add a service group
         if (is_null($user)) {
             throw new \Exception("Unregistered users can't create service groups.");
+        }
+
+        $hasAdminCredentials = $user->isAdmin();
+        $roleService = \Factory::getRoleService();
+        $userRoles = $roleService->getUserRoles($user);
+
+        $isUserValid = $hasAdminCredentials ? true : !empty($userRoles);
+
+        if (!$isUserValid) {
+            throw new Exception(
+                "You do not have permission to add a new "
+                . "Service Group. To add a new Service Group, you require "
+                . "at least one role assigned over an entity in GOCDB."
+            );
         }
 
         // can user assign reserved scopes ?


htdocs/web_portal/controllers/service_group/add_service_group.php
```diff
@@ -67,16 +69,30 @@ function submit($user) {
...
         }
 
+        $roleService = \Factory::getRoleService();
+        $hasRolesAssigned = $roleService->getUserRoles($user);
+
+        $isUserValid = $hasAdminCredentials ? true : !empty($hasRolesAssigned);
+
+        if (!$isUserValid) {
+            throw new Exception("You will need to have at least one role"
+                    . " assigned for adding new service group.");
```
rowan04, line 90:
```suggestion
                    . " assigned for adding a new service group.");
```

htdocs/web_portal/controllers/service_group/add_service_group.php
```diff
@@ -67,16 +69,30 @@ function submit($user) {
     }
 }
 
-function draw($user) {
+function draw($user)
+{
     try {
-        // If the user is registered they're allowed to add a service group
```
rowan04, line 72: as this comment is getting removed, maybe some new comments need to be added

Sae126V: I think variable names are giving the enough information the developer needs. So, I don't think a new comment is needed here. Does that sounds good?

htdocs/web_portal/controllers/service_group/add_service_group.php
```diff
@@ -67,16 +69,30 @@ function submit($user) {
...
+        $hasAdminCredentials = $user->isAdmin();
+
         if (is_null($user)) {
             throw new \Exception("Unregistered users can't create service groups.");
         }
 
+        $roleService = \Factory::getRoleService();
+        $hasRolesAssigned = $roleService->getUserRoles($user);
+
+        $isUserValid = $hasAdminCredentials ? true : !empty($hasRolesAssigned);
```
rowan04, line 86: what is actually happening on this line?

Sae126V: Case1: User is NOT allowed to add new SG's. If a User have no roles assigned.
Case2: User is allowed to add new SG's. If a user is registered and have at least one role assigned with 'STATUS_GRANTED'.

(OR)
 
If a user is an Admin -> they are allowed to ADD new service group's.

That code can help to decide `isUserValid` to perform the task we need. In our case that would be case 1 and case 2.

rowan04: i was more wondering how the " ? true : " works - i'm not doubting it doesn't work, just wondering how

Sae126V: Cool. So, I am using ternary operator here. The way it works is "variableName = conditionToEvaluate ? trueBlock : falseBlock". It is a shorthand notation for if...else  like
`
if (conditionToEvaluate) {
    trueBlock
} else {
    falseBlock
}
`
Does that make sense now ?

rowan04: ah right, thanks

rowan04 (commented):
lgtm i think, but Greg or someone else should give it a look over before merging

htdocs/web_portal/controllers/service_group/add_service_group.php
```diff
@@ -67,16 +69,30 @@ function submit($user) {
...
+{
     try {
-        // If the user is registered they're allowed to add a service group
+        $hasRolesAssigned = false;
+        $isUserValid = false;
+        $hasAdminCredentials = $user->isAdmin();
+
         if (is_null($user)) {
             throw new \Exception("Unregistered users can't create service groups.");
         }
```
gregcorbett, line 82: Switching the order of these is probably safer, as there's no guarantee `$user` is not `NULL` when the isAdmin method gets called (though it does seem PHP will happily have things like `NULL->someMethod()` return `NULL`).

```suggestion
        if (is_null($user)) {
            throw new \Exception("Unregistered users can't create service groups.");
        }

        $hasRolesAssigned = false;
        $isUserValid = false;
        $hasAdminCredentials = $user->isAdmin();
```

htdocs/web_portal/controllers/service_group/add_service_group.php
```diff
@@ -67,16 +69,30 @@ function submit($user) {
...
 
+        $roleService = \Factory::getRoleService();
+        $hasRolesAssigned = $roleService->getUserRoles($user);
+
+        $isUserValid = $hasAdminCredentials ? true : !empty($hasRolesAssigned);
+
+        if (!$isUserValid) {
+            throw new Exception("You will need to have at least one role"
+                    . " assigned for adding a new service group.");
```
gregcorbett, line 90:
```suggestion
            throw new Exception("You do not have permission to add a new"
                    . " ServiceGroup. To add a new ServiceGroup you require"
                    . " a role over an existing GOCDB enitity.");
```

gregcorbett (changes requested):
two minor changes :)

htdocs/web_portal/controllers/service_group/add_service_group.php
Sae126V: Sure, Greg. Updated the exception message with you suggestion.

htdocs/web_portal/controllers/service_group/add_service_group.php
Sae126V: Updated the code.

htdocs/web_portal/controllers/service_group/add_service_group.php
```diff
@@ -67,16 +69,31 @@ function submit($user) {
...
+        $hasAdminCredentials = $user->isAdmin();
+        $roleService = \Factory::getRoleService();
+        $hasRolesAssigned = $roleService->getUserRoles($user);
+
+        $isUserValid = $hasAdminCredentials ? true : !empty($hasRolesAssigned);
+
+        if (!$isUserValid) {
+            throw new Exception("You do not have permission to add a new"
+                    . " ServiceGroup. To add a new ServiceGroup you require"
+                    . " a role over an existing GOCDB enitity.");
```
rowan04, line 91:
```suggestion
                    . " a role over an existing GOCDB entity.");
```

rowan04 (commented):
Apart from what i just reviewed, lgtm

Sae126V:
Ready to be reviewed.

gregcorbett:
Looks good to me, can you remove "(F)" from the commit message please?

Sae126V:
> Looks good to me, can you remove "(F)" from the commit message please?

Done. I have reworded the commit. The reason I have added is I wanted to differentiate styles vs functionality.

gregcorbett:
rebasing on top of latest dev

PR #467, opened 2023-08-14
[GT-193] Update css properties for service group

### LINKED ISSUE #374: Layout can be wonky on ServiceGroup page
I'm not sure if it's caused by a ServiceGroup with a particularly long name, a particularly long description or something else all together - but it's possible the layout can be wonky for a ServiceGroup.

For example: https://goc.egi.eu/portal/index.php?Page_Type=Service_Group&id=3233

image

Looking at that example, I suspect it's an issue with the length of the description, maybe we can force some line breaks?

Sae126V:
Update CSS properties to support service group names being displayed properly. So that the long sentence(words) can be broken down based on the parent `div`.

Resolves #374

### DIFF
--- a/htdocs/web_portal/views/service_group/view_sgroup.php
+++ b/htdocs/web_portal/views/service_group/view_sgroup.php
@@ -7,22 +7,39 @@ $showPD = $params['authenticated'];
 
 <script type="text/javascript" src="<?php echo \GocContextPath::getPath()?>javascript/confirm.js"></script>
 <div class="rightPageContainer">
-    <div style="float: left;">
-        <img src="<?php echo \GocContextPath::getPath()?>img/virtualSite.png" class="pageLogo" />
-    </div>
-    <div style="float: left;">
-        <h1 style="float: left; margin-left: 0em;">
-                Service Group: <?php xecho($params['sGroup']->getName())?>
-        </h1>
-        <span style="clear: both; float: left; padding-bottom: 0.4em;"><?php xecho($params['sGroup']->getDescription())?></span>
-    </div>
+    <div style="display: flex; justify-content: space-between; gap:10%">
+        <div style="display:flex; flex-direction: row;">
+            <div>
+                <img
+                    src="<?php
+                        echo \GocContextPath::getPath();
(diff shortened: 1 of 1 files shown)


gregcorbett:
ah so the desired behaviour here is for the icon and site name to stay where they are, and the description text wrap if it needs to, similar to what happens for Sites with a long description:

image

Sae126V:
Updated the code to implement site like functionality, Greg. Done

htdocs/web_portal/views/service_group/view_sgroup.php
```diff
@@ -7,30 +7,55 @@
...
-        <h1 style="float: left; margin-left: 0em;">
-                Service Group: <?php xecho($params['sGroup']->getName())?>
-        </h1>
-        <span style="clear: both; float: left; padding-bottom: 0.4em;"><?php xecho($params['sGroup']->getDescription())?></span>
-    </div>
+    <div
+        style="display: flex;
+            justify-content: space-between;
+            gap:10%"
+    >
```
gregcorbett, line 14: Would this fit (just) on a single line 80 character line?

htdocs/web_portal/views/service_group/view_sgroup.php
```diff
@@ -7,30 +7,55 @@
...
+            <div style="display:flex; flex-direction: row;">
```
gregcorbett, line 15: I think this is slightly over indented compared to the above line?

gregcorbett (commented):
Functionally this works 😄, I've left a couple of questions about line splitting and indentation

htdocs/web_portal/views/service_group/view_sgroup.php
Sae126V: Ah right. Updated the code in eae6644c0ac151f6a1840ac81e274a0c22e99735

gregcorbett:
LGTM, can you squash the commits into the first one?

gregcorbett (approved):
LGTM :)

gregcorbett:
rebasing on top of latest dev

PR #468, opened 2023-08-14
Add a check on line length to PHP Codesniffer

gregcorbett:
According to PSR-12, lines SHOULD NOT be longer than 80 characters, so I think it would be good to flag if lines go over 80 so we can see if they can easily be broken down.


gregcorbett:
> According to PSR-12, lines SHOULD NOT be longer than 80 characters, so I think it would be good to flag if lines go over 80 so we can see if they can easily be broken down.

80 also works well on smaller laptop screens, enabling a split diff to be disaplayed in GitHub without wrapping.

Sae126V:
LGTM;

Sae126V (approved):

PR #470, opened 2023-08-15
[GT-145] Fixing, Refreshing and Enriching the Service Group Sample Data

### LINKED ISSUE #416: Refresh and enrich ServiceGroups from sample data
Follows on from #394 

- [ ] Extend `lib/Doctrine/deploy/AddServiceGroups.php` so that it can create ServiceGroups that actually contain Services
- [ ] Add `lib/Doctrine/deploy/AddServiceGroupRoles.php` so that ServiceGroup level roles can be created.
- [ ] Refresh `lib/Doctrine/deploy/sampleData/ServiceGroups.xml` and `/lib/Doctrine/deploy/sampleData/ServiceGroupRoles.xml` to be consistent with the rest of the sample data.

The best way to tackle this may be to deploy the existing sample data to an instance, create some ServiceGroups in the UI, then use the [API output](https://gocdb.github.io/api/read/get_service_group/) to update the `ServiceGroups.xml`. Because the GOCDB UI doesn't let you request a role on behalf of a user, `ServiceGroupRoles.xml` can start from the [API output](https://gocdb.github.io/api/read/get_service_group_role/) but then it will have to be manually updated to include user roles.

rowan04:
will resolve #416 when finished


rowan04:
rebased to remove extra commit

rowan04:
to note - all service groups are currently incorrectly given only local scope

rowan04:
> to note - all service groups are currently incorrectly given only local scope

fixed :)

rowan04:
Codacy and codeclimate issues to be fixed. Other than that, ready for review

rowan04:
force push and rebase to merge 3 commits on code formatting fixes into 1 commit.
All codacy and codeclimate issues are now fixed.

lib/Doctrine/deploy/AddServiceGroupRoles.php
```diff
@@ -0,0 +1,93 @@
...
+         * one role type
+         */
+        if (count($roleTypes) !== 1) {
+            throw new Exception(count($roleTypes) . " role types " .
+                "found with name: " . $userRole);
+        }
+
+        foreach ($roleTypes as $result) {
+            $roleType = $result;
+        }
```
gregcorbett, line 40: What is this (and similar) for loop doing?

gregcorbett (commented):
Looks good, just one comment.

lib/Doctrine/deploy/AddServiceGroupRoles.php
rowan04: These loops are present in similar files such as https://github.com/GOCDB/gocdb/blob/dev/lib/Doctrine/deploy/AddNGIRoles.php from which I copied them over. Deploying the sample data doesn't work without them, and I think they are there as the corresponding database query above returns an array, but only with one item in it. The foreach loop takes the one data item in the array and makes it a normal string, which is then used later on to make the role in the Role() function.

gregcorbett: ah okay - I think something like `$roleType = $roleTypes[0]` would be clearer as to what is happening. Could you change that for AddServiceGroupRoles.php?

rowan04: ok

rowan04: done

gregcorbett (approved):

gregcorbett:
rebasing on top of latest dev

PR #479, opened 2023-09-13
[GT-201] Give sample data sites their correct timezone, instead of UTC by default

### LINKED ISSUE #469: SampleData deploy scripts give sites the UTC timezone by default
Despite the fact each sample data site has its own timezone set, the deploy script doesn't use it and just sets each site's timezone to UTC.

rowan04:
- change to use the timezone provided in the site xml, instead of UTC by default

resolves #469


gregcorbett (approved):
LGTM

PR #483, opened 2023-09-15
[GT-183] Apply code styles for the files #461

Sae126V:
Applied code styles for the files touched in #461 .

Applied:
1) Indentation
2) Used concatenation if the line exceeds the limit(Except for the comments already exists) 

- Provided  a blank line between the variables and conditions.
- For Return
      - If it is the single statement in the block (Not provided a blank line).
      - In any other case always provided a blank line before it.
 - Provided a blank line between div blocks in HTML.
 - Provided a blank line for element in HTML (Optional - Depends on where we are using it).


Sae126V:
Ready to be reviewed for the code issues

htdocs/web_portal/controllers/service_type/view_service_type.php
```diff
@@ -20,12 +21,14 @@
  * limitations under the License.
  /*======================================================*/
 require_once __DIR__ . '/../utils.php';
-require_once __DIR__ . '/../../../web_portal/components/Get_User_Principle.php';
+require_once __DIR__
+    . '/../../../web_portal/components/Get_User_Principle.php';
```
tofu-rocketry, line 25: Is it possible to use parentheses around `__DIR__ . '/../../../web_portal/components/Get_User_Principle.php'` to make it clear this is the same argument to `require_once`.

(Note, I am not a PHP expert! I don't know if that is a good idea or not!)

htdocs/web_portal/views/admin/edited_service_type.php
```diff
@@ -9,8 +9,13 @@
...
-        Click here</a> to edit the <?php xecho($params['Name'])?> service type again.
-
+        <a
+            href="index.php?Page_Type=Admin_Edit_Service_Type&amp;id=<?php
+                echo $params['ID'];
+            ?>"
+        >
+        Click here</a>to edit the <?php
+            xecho($params['Name']);
+        ?> service type again.
```
tofu-rocketry, line 19: This feels a bit too broken up and harder to understand.

PR #464, opened 2023-08-04
[GT-184] Add support for renewing API credentials

Sae126V:
Resolves 

" #438 and #453 "

PR #484, opened 2023-09-19
[GT-184] Apply styles for the PR #464

Sae126V:
In extend to #464 . Adding the styles PR with few minor improvements.


htdocs/web_portal/views/site/view_site.php
```diff
@@ -12,174 +13,270 @@
...
+            <a
+                onclick="return confirmSubmit()"
+                href="index.php?Page_Type=Delete_Site&id=<?php
+                    echo($entityId);
+                ?>"
+            >
+                <img
+                    src="<?= \GocContextPath::getPath(); ?>img/trash.png"
+                    class="trash"
+                />
```
tofu-rocketry, line 53: These split tags are a bit ugly, but not sure what the better solution is.

PR #492, opened 2023-10-12
Update CURLOPT_SSL_VERIFYHOST

gregcorbett:
- CURLOPT_SSL_VERIFYHOST no longer accepts the value 1 (true), and hasn't for a while:tm:.
- value 2 was being implicity used instead.
- This change reduces PHP Notices in the error log.


tofu-rocketry (approved):
Found it useful to read the option's docs to follow along: https://curl.se/libcurl/c/CURLOPT_SSL_VERIFYHOST.html

PR #493, opened 2023-10-12
Rework Home Site display

gregcorbett:
Having the line break in the href string caused the spaces to be interpreted as part of the URL, which led to URLs like:

> portal/index.php?Page_Type=Site&id=%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20 5


htdocs/web_portal/views/user/view_user.php
```diff
@@ -122,10 +122,12 @@
                     <tr class="site_table_row_2">
                         <td class="site_table">Home Site</td>
                         <td class="site_table">
-                            <a href="index.php?Page_Type=Site&amp;id=
-                                <?php echo $params['user']->getHomeSite()->getId()?>">
-                                <?php xecho($params['user']->getHomeSite()->getShortName()) ?>
```
tofu-rocketry, line 127: What was `xecho` doing such that we can now use a plain `echo`?

htdocs/web_portal/views/user/view_user.php
```diff
@@ -122,10 +122,12 @@
...
-                            </a>
+                            <?php
+                            $homeId = $params['user']->getHomeSite()->getId();
+                            echo "<a href=\"index.php?Page_Type=Site&amp;id=$homeId\">";
+                            echo $params['user']->getHomeSite()->getShortName();
```
tofu-rocketry, line 128: As you're using a variable on line 127, I feel this part would be easier to read if it also used a variable that was set between 126 and 127.

htdocs/web_portal/views/user/view_user.php
```diff
@@ -122,10 +122,12 @@
                     <tr class="site_table_row_2">
                         <td class="site_table">Home Site</td>
                         <td class="site_table">
-                            <a href="index.php?Page_Type=Site&amp;id=
-                                <?php echo $params['user']->getHomeSite()->getId()?>">
-                                <?php xecho($params['user']->getHomeSite()->getShortName()) ?>
-                            </a>
+                            <?php
+                            $homeId = $params['user']->getHomeSite()->getId();
```
tofu-rocketry, line 126: Does seem so from other bits in the file, but to check, is this the correct indentation for this kind of block?

htdocs/web_portal/views/user/view_user.php
gregcorbett: ah good spot, I should use `xecho` as the short name is user submitted input. Fixed in e074ed612e089867a38ad22cc0923d17e8740864.

htdocs/web_portal/views/user/view_user.php
gregcorbett: yep, that is to say, a `<?php` tag doesn't increase the level of indentation.

gregcorbett:
post review, I will squash the 5 commits into one.

tofu-rocketry (approved):

PR #494, opened 2023-10-18
Drop PHP 5.X support

gregcorbett:
~Rebase after #509 is merged.~

This PR removes `composer-5.4.{json, lock}` to `composer.{json.lock}` and drops PHP 5.x from the CI tests.


tofu-rocketry (commented):
Why are are the other 7.xs dropped (no mention in the commit)?
Are there no later versions you want to test against? (Or does it just not work with any of the 8.x series?)

gregcorbett:
> Why are are the other 7.xs dropped (no mention in the commit)?

Fair point, for completeness - it's because 7.4 is the version of PHP 7 readily install-able on Rocky8 (see somewhat obscure reference in e2042aa7aa15bc0501ca2ccf2ad8717ea599a637)

> Are there no later versions you want to test against? (Or does it just not work with any of the 8.x series?)

8.X requires a later release of version 2.X of the Doctrine ORM before the tests will even run (279606c1ea84c3b6153f41ffccb8b592ce648891)

PR #499, opened 2023-11-23
 Update ShibToken for Keycloak based EGI Check-in to master

gregcorbett:
This has included:
  - switching to aai-demo.egi.eu for preproduction, on the advice of EGI Check In team
  - updating the expected IdPs from `subdomain/proxy/saml2/idp/metadata.php` to `subdomain/auth/realms/egi`
  - moving from using epuid to voPersonID

Related ticket: https://ggus.eu/index.php?mode=ticket_info&ticket_id=164026

Once this is confirmed to work in production, before merging I'll bump the version number up to 5.10.5. This PR deliberately merges into master so that it can be released quckly, in isolation.


Sae126V:
Just out of curious, Greg. So for this client in IRIS IAM, the client has selected keyclock profile so that the gocdb receives the information (when exchanged) Say `groups` to `roles` claim etc ?

PR #509, opened 2024-02-27
Soft drop PHP 5.X support, in favour of PHP 7.4 "out of the box"

gregcorbett:
This PR soft drops support for PHP 5.X, in favour of PHP 7.4 support "out of the box".

It is still possible to run GOCDB on PHP 5.X - however this will require specific deployments reverting this commit to restore the appropriate composer files. 

The `composer-5.4.{json, lock}` files will remain in the short term to facilitiate the above and to enable continued testing of the codebase against PHP 5.X. These files will be removed from the repository after 1st July 2024, along with automated testing of PHP 5.X.

NB: The diff rendered here is slightly weird...
- it appears as a rename of the composer-7.4 files to the composer-5.4 files, which is not what has happened.
- the pre-existing composer files were renamed to composer-5.4 files and then the composer-7.4 files were renamed to the composer files.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -18,17 +18,17 @@ jobs:
         extension: ["pdo_mysql"]
         include:
           - php-version: "5.4"
-            composer-json: "composer.json"
+            composer-json: "composer-5.4.json"
           - php-version: "5.5"
-            composer-json: "composer.json"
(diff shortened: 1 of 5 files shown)


gregcorbett:
#494 is the PR to actually drop PHP 5.X support.

Sae126V (approved):
LGTM.

PR #525, opened 2024-09-12
Update to ORM 2.10

gregcorbett:
ORM 2.10 seems to be as high as we can go now without starting to hit errors that at first glance seem to be because of our version of PHPUnit.

> Exception: Serialization of 'Closure' is not allowed

NB: A workable set of packages under ORM 2.11 was once achievable with our version of PHPUnit in https://github.com/GOCDB/gocdb/pull/455, but I guess the ecosystem has move on since then - using the commits in https://github.com/GOCDB/gocdb/pull/455 would result in some packages being downgraded from the current state of `dev`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "stfc-egi/gocdb",
     "description": "A web portal and REST style API for e-Infrastructure topology managment",
     "require": {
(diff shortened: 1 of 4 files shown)


gregcorbett:
See https://github.com/GOCDB/gocdb/issues/527

PR #526, opened 2024-09-12
Update to ORM 2.9

gregcorbett:
ORM 2.9 seems to be as high as we can go now without starting to hit errors that at first glance seem to be because of our version of PHPUnit.

> Exception: Serialization of 'Closure' is not allowed

NB: A workable set of packages under ORM 2.10 and 2.11 was once achievable with our version of PHPUnit in https://github.com/GOCDB/gocdb/pull/455, but I guess the ecosystem has move on since then - using the commits in https://github.com/GOCDB/gocdb/pull/455 would result in some packages being downgraded from the current state of dev.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "stfc-egi/gocdb",
     "description": "A web portal and REST style API for e-Infrastructure topology managment",
     "require": {
-	"doctrine/orm": "2.7.*"
+	"doctrine/orm": "2.8.*"
     },
     "require-dev": {
         "phpunit/phpunit": "4.8.*",
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,8 @@
     "name": "stfc-egi/gocdb",
     "description": "A web portal and REST style API for e-Infrastructure topology managment",
     "require": {
-	"doctrine/orm": "2.8.*"
+        "doctrine/orm": "2.9.*",
+        "symfony/cache": "*"
     },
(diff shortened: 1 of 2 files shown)


gregcorbett:
See https://github.com/GOCDB/gocdb/issues/527 for going to ORM 2.10

tofu-rocketry (commented):
You planning to squash this? Or at least the individual ORM updates?

What's the `symfony/cache` about?

Does the `composer.lock` actually need reviewing, or can we trust in the auto generation?

gregcorbett:
> You planning to squash this? Or at least the individual ORM updates?

I was going to leave them as is, that way if we need to roll back versions of ORM we have options other than back to 2.6
 
> What's the `symfony/cache` about?

Hmm good point... *goes digging* ah-hah! It was discussed https://github.com/GOCDB/gocdb/pull/344#discussion_r869305680, but the resultant commit message made it's way to the ORM 2.10 commit.

I'll update the commit message, and fix the indentation, but basically this 😋: https://github.com/doctrine/orm/blob/2.9.x/UPGRADE.md#minor-bc-break-setup-tool-needs-cache-implementation

> Does the `composer.lock` actually need reviewing, or can we trust in the auto generation?

I am happy to trust the auto generation 😆

gregcorbett:
> and fix the indentation

ah it's tabs and space 😋

tofu-rocketry (approved):

Issue #377, opened 2022-08-11
Rename Link_Identity page along with associated views and controllers


gregcorbett:
As they also handle account recovery, calling them Link_Identity (i.e https://goc.egi.eu/portal/index.php?Page_Type=Link_Identity) or similar is confusing. Maybe they should be called something like "Identifier_Management".

associated views and controllers: 
- htdocs/web_portal/views/user/
  - link_identity.php
  - link_identity_accepted.php
  - link_identity_rejected.php
  - linked_identity.php 
- htdocs/web_portal/controllers/user/
  - link_identity.php
  - link_identity_user_validate.php

gregcorbett:
related [GGUS Ticket](https://ggus.eu/index.php?mode=ticket_info&ticket_id=158446).

gregcorbett:
related [EGI JIRA Issue](https://jira.egi.eu/browse/EGITCB-396).

tofu-rocketry:
Will go with `identifier_management`.

PR #528, opened 2024-10-16
Rename link Identity to Identifier Management

Sae126V:
Resolves [GT-579] and https://github.com/GOCDB/gocdb/issues/377


Sae126V:
Finished fixing the line length and indentation issues.

gregcorbett:
I've applied the schema change label as this renames an entity - so presumably we create new tables.

gregcorbett:
LGTM - I'll roll it out to a host for final testing and then merge 😄

PR #529, opened 2024-11-13
Add code to support admin ONLY x509 login for the portal

Sae126V:
- Code to support admin only x509 login. For rest of the users they are required to use other means.


gregcorbett:
The code looks good, I'll try and rework it so it's more authentication method agnostic.

gregcorbett:
rebasing to bring up to date

Issue #423, opened 2023-02-03
Encapsulate super-global _GET, _POST, etc into objects


gregcorbett:
If we can, we should use an existing framework to do this, such as the [Symfony HttpFoundation Component](https://symfony.com/doc/current/components/http_foundation.html).

gregcorbett:
This would make parts of GOCDB (such as the Authentication code) more testable.

PR #533, opened 2025-01-24
Improve handling when no certificate is provided.

gregcorbett:
PHP 7.4 (sensibly) raises a notice when trying to access `false` as if it where an array, i.e. via `$Plain_Client_Cerfificate['name']`

Accessing `false` as if it were an array is not enough to break anything currently - but the notice does get raised every time someone access GOCDB via, for example, EGI CheckIn.

This PR address that issue and eliminates the notice.


gregcorbett:
The codeclimate issues predate this change and are tracked in #423.

Sae126V (approved):
LGTM, Greg.

PR #534, opened 2025-01-29
Fix handling of missing X509 certificate

gregcorbett:
- $Plain_Client_Cerfificate is an array when a certificate has been provided, and a boolean (false) when no certificate is provided.
- As such, if _and only if_ $Plain_Client_Cerfificate _is not_ an array do we want to return here.


jrha (approved):
:laughing:

PR #535, opened 2025-02-28
Enhance check on database used on monitoring page

gregcorbett:
Resolves GT-759

Following updates to database infrastructure, new failure modes have presented themselves - i.e. the database is healthy enough to pass the basic connection check - but not healthy enough to be used beyond that.

This is a heavier weight (but by no means a heavy weight) check that actually tries to use the database and may be more useful here.

It is certainly no worse, i.e. if the database is down - this new check still fails.


gregcorbett:
I might rework this into a hot fix, and push it out to prod so we can have it running the next time the database goes funky

gregcorbett:
Reviewed in #536

PR #536, opened 2025-03-03
Enhance check on database used on monitoring page (master)

gregcorbett:
Resolves GT-759

Reworked branch into a hotfix as discussed in GT-759 and #535, as such it deliberately merges into `master`.

Following updates to database infrastructure, new failure modes have presented themselves - i.e. the database is healthy enough to pass the basic connection check - but not healthy enough to be used beyond that.

This is a heavier weight (but by no means a heavy weight) check that actually tries to use the database and may be more useful here.

It is certainly no worse, i.e. if the database is down - this new check still fails.

### DIFF
--- a/htdocs/web_portal/GOCDB_monitor/tests.php
+++ b/htdocs/web_portal/GOCDB_monitor/tests.php
@@ -136,8 +136,11 @@ function test_db_connection()
         $entityManager = Factory::getNewEntityManager();
         // Get the database connection.
         $connection = $entityManager->getConnection();
-        // Try a simple select statement.
-        $connection->fetchOne('SELECT 1');
+        // Try a simple select statement - attempt to bypass any caching
+        // by requesting the current timestamp to the nearest millisecond.
+        // This is done by the 3, which represents the number of decimal
+        // to display.
+        $connection->fetchOne('SELECT CURRENT_TIMESTAMP(3)');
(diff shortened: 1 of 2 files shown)


gregcorbett:
This is ready for review - but not merging as I will (tag and) deploy the release candidate into pre-production first.

Assuming that deployment doesn't raise any issues, I'll bump the version number and merge.

htdocs/web_portal/GOCDB_monitor/tests.php
```diff
@@ -134,7 +134,13 @@ function test_db_connection()
     $retval = [];
     try {
         $entityManager = Factory::getNewEntityManager();
-        $entityManager->getConnection()->connect();
```
tofu-rocketry, line 137: Is `connect` not needed?

DonaldChung-HK: Does it require `;` to close off the SQL statement?

gregcorbett: It does not 😄 https://www.doctrine-project.org/projects/doctrine-dbal/en/3.0/reference/data-retrieval-and-manipulation.html#fetchone

DonaldChung-HK (approved):
Looks good to me. Thanks for putting this together, @gregcorbett .

PR #548, opened 2025-08-06
Add scripts to ease migration to new WLCG hierarchical scope structure

gregcorbett:
Resolves GT-1074

Add scripts should automate creation and then application of the new scope tags.

WLCGScopeTagCreateRunner.php, desired behaviour: Create every combination of VO.tierN for the supplied WLCG VO and tierN scope tags.

WLCGScopeTagAddRunner.php, desired behaviour: Sites get every possible combination of their existing WLCG VO and tierN scope tags in the new hierarchical structure.

I'd particularly appreciate checks that the logic is correct and achieves the desired behaviour.

PR #550, opened 2025-09-08
Extend WLCG scope migration scripts to cover Services

gregcorbett:
As a follow on from #548.

Much like Sites, the desired end state for Services can be inferred from the current scope tag set.

This will save Site admin's going through and updating all their services by hand.


tofu-rocketry (approved):
Looks ok, other than PHP's own choice of full-stops for concatenate. 😅

